Issue #10, opened 2026-02-07
EPIC-10: UX Polish & Accessibility


steilerDev:
## Epic: UX Polish & Accessibility

**Epic ID**: EPIC-10
**Priority**: Could Have
**Sprint Suggestion**: Sprint 5

### Description
Final UX polish and usability enhancements that go beyond the basic functionality delivered in other epics. This covers cross-cutting usability concerns: comprehensive keyboard shortcuts across the entire application, advanced drag-and-drop refinements, touch optimization for mobile/tablet beyond basic responsiveness, and performance tuning to ensure smooth interactions. While each feature epic includes basic usability for its domain, this epic addresses the holistic, cross-cutting UX quality.

### Requirements Coverage
- **Section 3.4**: Usability
  - Intuitive interface (cross-cutting polish)
  - Minimal learning curve (onboarding/guidance)
  - Drag-and-drop for Gantt chart (advanced refinements beyond EPIC-06)
  - Keyboard shortcuts for power users (comprehensive, cross-cutting)
  - Touch friendly interface on mobile devices (advanced optimization)
- **Section 3.2**: Technical Requirements
  - Smooth Gantt chart interactions (performance tuning)
  - Fast load times (<2s) (performance optimization)

### Dependencies
- All other epics (this is a polish pass after core features are built)

### Goals
- Comprehensive keyboard shortcut system across all views
- Refined drag-and-drop interactions (not just Gantt -- also for reordering, status changes)
- Touch gesture optimization for tablet and mobile
- Performance profiling and optimization for smooth interactions
- Loading state indicators and skeleton screens
- Error handling UX (user-friendly error messages, recovery actions)
- Consistent visual design and interaction patterns across all features

PR #420, opened 2026-03-03
feat(paperless): add tag-based document auto-filtering (#419)

### LINKED ISSUE #419: 8.8: Paperless-ngx Tag-Based Document Auto-Filter
**As a** homeowner, **I want** to configure a Paperless-ngx tag name via an environment variable so that the document browser and all document searches automatically filter to only show documents with that tag, **so that** I only see project-relevant documents from my Paperless-ngx instance instead of my entire document archive.

**Parent Epic**: #8
**Priority**: Should Have

### Acceptance Criteria

- [ ] 1. A new optional environment variable `PAPERLESS_FILTER_TAG` is recognized by the server config plugin. When not set or empty, behavior is unchanged (all documents shown). When set, its value is treated as a Paperless-ngx tag **name** (string, case-sensitive).
- [ ] 2. On server startup, if `PAPERLESS_FILTER_TAG` is set and Paperless-ngx is enabled, the server resolves the tag name to its numeric Paperless-ngx tag ID by calling the Paperless-ngx tags API. If the tag name cannot be found, the server logs a warning but continues to operate normally (no crash, no filter applied — all documents shown). The resolved tag ID is cached for the lifetime of the server process.
- [ ] 3. When a filter tag is resolved, the `GET /api/paperless/documents` endpoint automatically appends the resolved tag ID to the `tags__id__in` query parameter sent to Paperless-ngx, in addition to any user-selected tag filters. This means documents must have BOTH the configured filter tag AND any user-selected tags.
- [ ] 4. When a filter tag is resolved, full-text search queries (`GET /api/paperless/documents?query=...`) also include the tag filter, so search results are scoped to the configured tag.
- [ ] 5. The `GET /api/paperless/documents/:id` (single document detail) endpoint is NOT filtered — any document accessible in Paperless-ngx can still be viewed by ID (this preserves existing document links that may reference documents outside the filter tag).
- [ ] 6. The `GET /api/paperless/status` response includes a new optional field `filterTag: { name: string, id: number } | null` indicating whether tag filtering is active and which tag is being used. When `PAPERLESS_FILTER_TAG` is not set, the field is `null`. When the tag name could not be resolved, the field is `{ name: "<configured-name>", id: null }` to indicate a misconfiguration.
- [ ] 7. The `GET /api/paperless/tags` endpoint is NOT filtered — all tags from Paperless-ngx are still returned (users need to see all tags for manual filtering within the already-filtered set).
- [ ] 8. The Document Browser UI displays a non-intrusive indicator (e.g., a small info badge or text near the search bar) when tag filtering is active, showing the filter tag name. This helps users understand why they may see fewer documents than expected.
- [ ] 9. The `PAPERLESS_FILTER_TAG` value is logged (not the resolved ID) at startup in the config log line when the variable is set, for debugging purposes.
- [ ] 10. The `AppConfig` interface is extended with an optional `paperlessFilterTag?: string` field. The `PaperlessStatusResponse` type is extended with the `filterTag` field described in AC #6.
- [ ] 11. The `PAPERLESS_FILTER_TAG` environment variable is documented in the project's environment variables table (CLAUDE.md, docs site, and docker-compose.yml).

### Notes

- This story extends the existing Paperless-ngx integration (EPIC-08) with server-side filtering
- The filter is applied server-side in `paperlessService.ts` — no client-side filtering needed
- The existing `tags__id__in` Paperless-ngx API parameter supports comma-separated tag IDs, so the filter tag ID can be prepended/appended to any user-selected tags
- Tag name resolution should happen once at startup (or on first request) and be cached, not on every request
- Edge case: if the filter tag is deleted from Paperless-ngx after server startup, document list requests will return 0 results (Paperless-ngx filters by non-existent tag). The status endpoint should still report the configured tag name. A server restart would re-resolve (and warn if not found).
- The env var uses a tag **name** (not ID) for better UX — users know their tag names but may not know internal IDs
- Dependencies: None (builds on completed EPIC-08 infrastructure)

steilerDev:
## Summary

- Add optional `PAPERLESS_FILTER_TAG` environment variable for automatic document pre-filtering
- When set, resolves tag name to Paperless-ngx tag ID and injects it into document list/search queries
- Status endpoint exposes `filterTag` field for UI awareness
- Document Browser shows info banner when tag filtering is active
- Graceful degradation when tag name not found (warning log, all documents shown)

Fixes #419

## Test plan
- [ ] Unit tests pass (config, service, routes, component)
- [ ] Integration tests pass

Co-Authored-By: Claude backend-developer (Haiku 4.5) <noreply@anthropic.com>
Co-Authored-By: Claude qa-integration-tester (Haiku 4.5) <noreply@anthropic.com>


steilerDev (commented):
**[product-architect]**

## Architecture Review: PR #420 — Tag-Based Document Auto-Filtering

### Critical Issues

**1. PR targets `main` instead of `beta`** (Process Violation)
Per project conventions, all feature PRs must target `beta`. This PR targets `main` with 21 commits spanning EPIC-04 household items, dependency bumps, and the filter tag feature. The branch must be rebased onto `beta` and the PR retargeted. As-is, this would bypass the integration branch entirely.

**2. Wiki API Contract not updated** (Contract Violation)
The `GET /api/paperless/status` response now includes two new fields (`paperlessUrl` and `filterTag`), but the wiki API Contract page still shows the old response shape without these fields. The wiki environment variables table also lacks `PAPERLESS_EXTERNAL_URL` and `PAPERLESS_FILTER_TAG`. These wiki updates must be part of this PR.

**3. `filterTag` response shape deviates from acceptance criteria** (AC #6 Deviation)
AC #6 specifies: `filterTag: { name: string, id: number } | null`. The implementation uses `filterTag: string | null` (just the tag name). While the simpler shape is arguably sufficient for the UI (which only needs the name for display), this is an undocumented deviation from the acceptance criteria. If the simpler shape is intentional, the issue AC should be updated to match. Either way, the wiki must document what was actually shipped.

### Medium Issues

**4. `docker-compose.yml` missing `PAPERLESS_FILTER_TAG` comment**
The docker-compose.yml includes a commented-out example for `PAPERLESS_EXTERNAL_URL` but omits `PAPERLESS_FILTER_TAG`. AC #11 requires documentation in docker-compose.yml. Add a commented example:
```yaml
#   - PAPERLESS_FILTER_TAG=cornerstone                  # Optional: only show documents with this tag
```

**5. Module-level cache has no invalidation path for tag deletion**
The `resolvedFilterTagId` cache lives for the process lifetime. If the filter tag is deleted from Paperless-ngx, `listDocuments` will continue filtering by a non-existent tag ID, returning 0 results with no way to recover short of restarting the server. The issue notes acknowledge this edge case, but the `getStatus` endpoint also caches a successful result and never re-checks. Consider adding a TTL or periodic re-validation on the status endpoint (even a long TTL like 1 hour would help). This is not a blocker but worth noting.

### Low Issues

**6. Case-insensitive tag matching not documented**
`resolveFilterTagId` does case-insensitive matching (`toLowerCase()`), but the issue AC #1 says "case-sensitive". The case-insensitive approach is more user-friendly, but this deviation should be noted.

**7. `_resetFilterTagCache` uses underscore-prefixed export**
The leading underscore convention for test-only exports is fine pragmatically but is not a documented project convention. This is informational — no action needed unless a convention is established.

### What Looks Good

- Config plugin follows existing patterns well (`loadConfig`, URL validation for `PAPERLESS_EXTERNAL_URL`)
- `paperlessFilterTag` is correctly logged at startup (AC #9)
- Tag merging logic with deduplication via `Set` is clean
- Graceful degradation when tag not found (warn + show all) is correct
- Filter banner UI is minimal and appropriate (role="note" for accessibility)
- Test coverage is thorough: config tests, service unit tests, route integration tests, component tests
- Documentation updates (docs site, CLAUDE.md env table) are present
- `PAPERLESS_EXTERNAL_URL` feature (bundled in this PR) is well-implemented with proper URL validation

### Verdict

Requesting changes primarily for the `main` targeting (critical process issue) and the missing wiki updates. The implementation quality is solid once the contract and process issues are resolved.

steilerDev (commented):
**[ux-designer]**

Design review for the UI changes in PR #420 (tag-based document auto-filtering). The client/src/ scope is small — one new CSS block in `DocumentBrowser.module.css` and the corresponding JSX in `DocumentBrowser.tsx`, plus `filterTag: null` additions to error-state fallbacks.

---

## Token Adherence

**Low — one hardcoded font-weight value.**

`DocumentBrowser.module.css` line 49:
```css
.filterBannerTag {
  font-weight: 600;  /* hardcoded */
}
```

The design system defines `--font-weight-semibold: 600`. Adjacent files in the same directory (`LinkedDocumentsSection.module.css`) already use the token correctly. The existing `DocumentBrowser.module.css` has the same pattern for its other `font-weight` declarations (500, 700 also hardcoded), so this new rule is consistent with the pre-existing file style — but it extends a pre-existing inconsistency rather than correcting it.

Recommended correction:
```css
.filterBannerTag {
  font-weight: var(--font-weight-semibold);
}
```

All other token usage in the new block is correct: `--spacing-2`, `--spacing-3`, `--color-primary-bg`, `--color-primary`, `--radius-md`, `--font-size-sm`, `--color-primary-badge-text` — all proper Layer 2 semantic references.

## Dark Mode Correctness

**Pass.** Every color token in the new `.filterBanner` block switches correctly in `[data-theme="dark"]`:
- `--color-primary-bg` → `rgba(59,130,246,0.15)` (dark)
- `--color-primary` → `var(--color-blue-400)` (dark)
- `--color-primary-badge-text` → `var(--color-blue-300)` = `#93c5fd` (dark)

Contrast check: `--color-primary-badge-text` (`#93c5fd`) on `--color-primary-bg` (`rgba(59,130,246,0.15)` blended on dark surface ~`#1e2940`) yields approximately 6.5:1 — passes WCAG AA. Light mode: `--color-badge-text` (`#1e40af`) on `#dbeafe` — same token pair used for `.tagChipActive`, already verified as passing.

## Visual Consistency

**Pass.** The `.filterBanner` correctly mirrors the `.tagChipActive` token pair (`--color-primary-bg` / `--color-primary` border / `--color-primary-badge-text`), making it visually coherent with the tag filter strip below it. The informational blue framing is appropriate for a read-only system-set filter (versus user-toggleable tag chips).

## Responsive Behavior

**Pass.** No explicit responsive overrides needed — the banner is a single-line flex row that wraps gracefully. At narrow viewports it will reflow within the `--spacing-2`/`--spacing-3` padding budget. The `flex: 1` on `.filterBannerText` ensures the text fills available width correctly.

## Accessibility

**Low — `role="note"` is a landmark role and semantically heavy for an inline banner.**

```tsx
<div className={styles.filterBanner} role="note" aria-label="Active document filter">
```

`role="note"` is a WAI-ARIA landmark role, which means screen readers expose it in the landmarks list (like `<nav>`, `<main>`, `<aside>`). For a small inline informational banner that is not a page-level region, this creates unnecessary landmark noise.

Since the filter is server-configured and static (loaded once on mount — no dynamic update after render), `aria-live` is not needed. The existing `aria-label="Active document filter"` provides adequate labeling.

Preferred approach:
```tsx
<p className={styles.filterBanner}>
  Showing only documents tagged{' '}
  <strong className={styles.filterBannerTag}>{hook.status.filterTag}</strong>
</p>
```

Or if a `div` is kept for layout reasons, drop `role="note"` entirely — the banner text is self-describing and positioned in the normal reading flow immediately after the search bar.

## Animation / Transitions

**Pass.** No animation is applied to `.filterBanner`. The banner appears as part of the initial render once status loads; no transition is needed and no `prefers-reduced-motion` guard is required.

## `filterTag: null` Error State Additions

**Pass.** The `filterTag: null` additions in `usePaperless.ts` (line 64) and `LinkedDocumentsSection.tsx` (line 84) correctly keep the fallback status objects in shape with the updated `PaperlessStatusResponse` type. No visual changes; no design concerns.

## Tokens.css Changes

**Pass.** The new amber palette tokens (`--color-amber-100`, `--color-amber-300`, `--color-amber-800`) and HI status / Gantt semantic tokens are correctly structured:
- Layer 1 palette tokens use raw hex values as expected
- Layer 2 semantic tokens reference Layer 1 via `var()`
- Dark mode Layer 3 overrides for HI status badges use appropriate rgba() translucent fills and reference existing dark palette tokens (`--color-amber-300`, `--color-blue-300`, `--color-emerald-200/400`)

These are documented in agent memory from Story 4.3 and 4.10 specs — implementation matches what was specified.

---

## Summary

Two findings, both low severity:
1. `font-weight: 600` hardcoded instead of `var(--font-weight-semibold)` in `.filterBannerTag` (pre-existing file pattern, but worth correcting)
2. `role="note"` is semantically too heavy for an inline banner — prefer dropping the role or using a `<p>` element

Neither is a blocker. The core design intent (informational blue banner using primary token family, consistent with active tag chip styling, correct dark mode behavior) is sound.

steilerDev (commented):
**[product-owner]** PR #420 Review — Story 8.8: Paperless-ngx Tag-Based Document Auto-Filter

**Verdict: REQUEST CHANGES** (3 blocking issues)

## Acceptance Criteria Assessment

| AC | Status | Notes |
|----|--------|-------|
| #1 | **FAIL** | `PAPERLESS_FILTER_TAG` is recognized and optional -- correct. However, the AC specifies "case-sensitive" matching, but the implementation uses `t.name.toLowerCase() === tagName.toLowerCase()` (case-insensitive). This contradicts the spec. |
| #2 | PASS | Tag name resolved to numeric ID via tags API. Warning logged when not found. Server continues. Cached at module level. |
| #3 | PASS | `listDocuments()` merges filter tag ID into `tags__id__in` with proper deduplication via `Set`. |
| #4 | PASS | `query` param passes through alongside tag filter. |
| #5 | PASS | `getDocument/:id` endpoint unchanged -- no filter applied. |
| #6 | **FAIL** | The AC specifies `filterTag: { name: string, id: number } | null`, and when the tag cannot be resolved: `{ name: "<configured-name>", id: null }`. The implementation uses `filterTag: string | null` -- a flat string, not an object. When the tag is not found, it returns `null` instead of `{ name: "<configured-name>", id: null }`. This means the UI cannot distinguish between "not configured" and "configured but tag not found" (misconfiguration). |
| #7 | PASS | Tags endpoint unchanged. |
| #8 | PASS | Filter banner renders with `role="note"` and tag name. Styled appropriately. |
| #9 | PASS | Config log includes `paperlessFilterTag` in the logged config object at startup. |
| #10 | PASS (partial) | `AppConfig` extended with `paperlessFilterTag?: string`. `PaperlessStatusResponse` extended -- but with `filterTag: string | null` instead of the object shape from AC #6. |
| #11 | **FAIL** | `PAPERLESS_FILTER_TAG` is documented in CLAUDE.md env vars table and docs site, but is **missing from docker-compose.yml**. The docker-compose.yml patch only adds `PAPERLESS_EXTERNAL_URL`, not `PAPERLESS_FILTER_TAG`. |

## Blocking Issues

### 1. AC #6: filterTag response shape mismatch

The AC explicitly defines the shape as `filterTag: { name: string, id: number } | null` with a fallback of `{ name: "<configured-name>", id: null }` for misconfiguration. The implementation returns a flat `string | null`. This prevents the frontend from showing a "tag configured but not found" warning, which is the explicit purpose of the `{ name, id: null }` shape.

**Fix**: Change `PaperlessStatusResponse.filterTag` from `string | null` to `{ name: string; id: number | null } | null`. Update `paperlessService.getStatus()` and the route handler to return the object shape. When tag is configured but not found, return `{ name: configuredName, id: null }`.

### 2. AC #11: PAPERLESS_FILTER_TAG missing from docker-compose.yml

The docker-compose.yml commented environment section documents `PAPERLESS_URL`, `PAPERLESS_API_TOKEN`, and `PAPERLESS_EXTERNAL_URL` -- but not `PAPERLESS_FILTER_TAG`. All three places (CLAUDE.md, docs, docker-compose.yml) are required by the AC.

**Fix**: Add `#   - PAPERLESS_FILTER_TAG=your-tag-name` to the commented Paperless env block in `docker-compose.yml`.

### 3. AC #1: Case sensitivity contradiction

The AC specifies the tag name is "case-sensitive", but `resolveFilterTagId()` uses `.toLowerCase()` on both sides of the comparison. Either the AC or the implementation needs to change. Since the AC was written deliberately, the implementation should match (remove `toLowerCase()`).

**Fix**: Change `data.results.find((t) => t.name.toLowerCase() === tagName.toLowerCase())` to `data.results.find((t) => t.name === tagName)` in `paperlessService.ts`.

## Scope Concern

This PR bundles 28,000+ lines from EPIC-04 (Household Items) together with Story 8.8. The EPIC-04 changes include 3 database migrations (0010, 0011, 0012), full CRUD for household items, budget integration, dependency scheduling, Gantt/calendar integration, and dozens of new pages/components. This is a significant scope concern -- Story 8.8 is a focused feature that should be reviewable independently. The bundled EPIC-04 code was not part of Issue #419 and makes this PR extremely difficult to review in isolation.

I am reviewing only the Story 8.8 scope per the issue request. The EPIC-04 code will need its own review cycle through the proper story-by-story workflow.

## Non-Blocking Observations

1. **PAPERLESS_EXTERNAL_URL** is also added in this PR but is not part of Issue #419. It appears to be a useful feature (separating internal API URL from browser-facing URL) but should be tracked as its own story or refinement item.

2. **E2E page object deletion**: The `DocumentsPage.ts` POM was deleted entirely. If this is replaced by inline locators in the spec file, that is fine, but verify E2E coverage is not lost.

3. **3 E2E shards failed** (4, 9, 15). The `Merge E2E Reports` gate passed, but the shard failures should be investigated before merge.

4. The security-engineer's LOW finding about tag name length validation is reasonable for a future refinement.

PR #487, opened 2026-03-06
fix(scheduling): Model milestones as CPM nodes to correctly determine critical path

### LINKED ISSUE #484: Bug: Critical path not correctly determined when milestone is on the path
## Bug Report

**Severity**: High
**Area**: Scheduling Engine (CPM), Timeline/Gantt visualization
**Affected files**: `server/src/services/schedulingEngine.ts`, `server/src/services/timelineService.ts`, `client/src/components/GanttChart/GanttChart.tsx`

### Problem

The Critical Path Method (CPM) implementation in the scheduling engine does not correctly handle milestones when they lie on the critical path. Milestones are zero-duration markers that can have dependencies (work items contribute to milestones, and work items can depend on milestones). The critical path should trace through milestones just like any other node, but currently milestones are invisible to the critical path calculation.

### Root Cause

In `autoReschedule()` (lines 692-760 of `schedulingEngine.ts`), milestones are handled by **expanding** them into synthetic dependencies rather than modeling them as CPM nodes:

1. **Completed milestones**: A virtual work item (`__milestone_N`) is injected with `actualEndDate = completedAt`, and a synthetic FS dep is added from it to the dependent work item. These virtual IDs appear in `criticalPath` but are filtered out or meaningless to the Gantt chart (which expects real work item IDs).

2. **Incomplete milestones**: No node is created at all. Instead, synthetic FS deps are generated directly from each contributor work item to each dependent work item, bypassing the milestone entirely. The milestone never appears in the CPM graph, so it **cannot** be on the critical path.

This means:
- The `criticalPath` array in `ScheduleResponse` never contains milestone identifiers for incomplete milestones
- The Gantt chart's `criticalPathSet` (used to highlight bars and arrows) only contains work item IDs — milestone diamonds are never highlighted as critical
- When the true critical path passes through a milestone (e.g., A -> Milestone M -> B), the synthetic expansion creates deps A -> B directly, but if there are multiple contributors to M, the critical path computation may incorrectly identify non-critical contributor paths as critical (fan-out distortion)
- The `criticalPath` type is `string[]` of work item IDs only (`shared/src/types/schedule.ts:23`, `shared/src/types/timeline.ts:110`), with no provision for milestone identifiers

### Expected Behavior

- Milestones should be modeled as zero-duration nodes in the CPM graph (similar to how completed milestones already get virtual work items)
- The critical path calculation should trace through milestone nodes: if a milestone has zero total float, it should be marked as critical
- The `criticalPath` response should include milestone identifiers when milestones are on the critical path
- The Gantt chart should highlight milestone diamonds with the critical path border color when they are critical
- The timeline API should propagate milestone criticality to the frontend

### Actual Behavior

- Incomplete milestones are not nodes in the CPM graph — they are expanded into direct work-item-to-work-item synthetic dependencies
- Milestones never appear in the `criticalPath` array
- Milestone diamonds on the Gantt chart are never highlighted as critical, even when they are on the longest path through the project
- The fan-out expansion (each contributor -> each dependent) may distort float calculations for work items around the milestone

### Acceptance Criteria

- [ ] **AC-1**: Given a linear dependency chain A (5 days) -> Milestone M -> B (3 days) where A and B are the only work items, when the scheduler computes the critical path, then the critical path contains A, M, and B (all three have zero total float)
- [ ] **AC-2**: Given a milestone M with two contributors (A: 5 days, C: 3 days) and one dependent (B: 2 days), when the scheduler runs, then the critical path includes A and M and B (the longest path through M), and C is not on the critical path (it has positive float)
- [ ] **AC-3**: Given a milestone on the critical path, when the GET /api/timeline endpoint returns data, then the milestone's `isCritical` property (or equivalent) is `true` in the response
- [ ] **AC-4**: Given a milestone on the critical path, when the Gantt chart renders with "Highlight Critical Path" enabled, then the milestone diamond marker is rendered with the critical path border/highlight color
- [ ] **AC-5**: Given a milestone that is NOT on the critical path (has positive float), when the Gantt chart renders with critical path highlighting, then the milestone diamond is rendered in its default style (no critical highlight)
- [ ] **AC-6**: Given a completed milestone (with `completedAt` set) on what was the critical path, when the scheduler runs, then the virtual milestone node is correctly included in the critical path if it has zero total float
- [ ] **AC-7**: Given multiple milestones in series (A -> M1 -> B -> M2 -> C), when the scheduler computes the critical path, then both milestones M1 and M2 appear on the critical path along with work items A, B, and C

### UAT Scenarios

**Scenario 1: Linear chain through milestone**
- **Given** work item A (5 days, not_started) with a successor milestone M, and work item B (3 days, not_started) that depends on milestone M
- **When** the timeline page loads with critical path highlighting enabled
- **Then** A's bar, M's diamond, and B's bar are all highlighted as critical path items

**Scenario 2: Milestone with multiple contributors**
- **Given** milestone M with contributors A (10 days) and C (3 days), and dependent B (5 days)
- **When** the critical path is computed
- **Then** A -> M -> B is the critical path; C has positive float and is not highlighted

**Scenario 3: Non-critical milestone**
- **Given** a milestone M that is not on the longest path (all contributors and dependents have float)
- **When** the Gantt chart renders
- **Then** M's diamond is rendered in default style, not critical path style

### Notes

- The `ScheduleResponse.criticalPath` type (`string[]`) may need to be extended or a parallel `criticalMilestones` field added to avoid breaking the existing API contract
- The `TimelineMilestone` type may need an `isCritical` boolean field
- The pure `schedule()` function needs to be able to receive milestone nodes (as zero-duration items) and include them in the critical path output
- This bug affects both the scheduling correctness (float calculations around milestones) and the visual representation (Gantt highlighting)
- Related code: `autoReschedule()` milestone expansion (lines 692-760), `getTimeline()` milestone section (line 273+), `GanttChart.tsx` critical path set (line 466)

steilerDev:
## Summary

Milestones were not appearing on the critical path, even when they sat on the longest path through the project. This fix models each milestone as a zero-duration CPM node so they are naturally included in critical path calculations.

**Root cause**: The previous implementation expanded milestones into synthetic work item→work item dependencies, causing milestones to disappear from the CPM graph entirely.

**Solution**: Model each milestone as a zero-duration CPM node with ID prefix `milestone:<id>`. The CPM engine naturally computes float for these nodes and includes them in the critical path when appropriate.

### Shared Types (`shared/src/types/timeline.ts`)
- Add `isCritical?: boolean` to `TimelineMilestone` interface

### Scheduling Engine (`server/src/services/schedulingEngine.ts`)
- **Section 3-4**: Replace milestone dependency expansion with milestone-as-CPM-node logic
  - Create one zero-duration `SchedulingWorkItem` per milestone with contributors or dependents
  - Construct milestone node ID as `milestone:<id>`
  - For completed milestones: use `completedAt` date; for incomplete: use `targetDate`
  - Create FS dependencies: each contributor→milestone and each milestone→dependent
  
- **Section 7**: Skip writing milestone nodes back to database
  - Filter out IDs starting with `milestone:` when writing scheduled dates
  - These are virtual CPM nodes only; they should never be persisted

### Timeline Service (`server/src/services/timelineService.ts`)
- **Section 5**: Pass milestone CPM nodes to the scheduling engine
  - Build milestone nodes using same logic as `autoReschedule()`
  - Build milestone synthetic dependencies
  - Add both to the engine's work items and dependencies
  - Extract critical milestone IDs from CPM result
  - Filter `milestone:` entries from returned `criticalPath` (API should only contain work item IDs)

- **Section 6**: Propagate `isCritical` to milestones
  - Add `isCritical` field to each `TimelineMilestone` in the response based on critical milestone IDs
  - Skip milestone nodes when building work item end date map for projectedDate computation

## Test Coverage

The QA agent will verify:
- CPM correctly identifies milestones on the critical path
- Non-critical milestones have `isCritical: false`
- Milestone nodes do not appear in the `criticalPath` array returned by the API
- `autoReschedule()` handles milestone nodes correctly during date updates

Fixes #484


steilerDev (commented):
**[security-engineer]**

PR #487 reviewed — no security findings.

**Scope**: Milestone CPM node refactor (, , , , 44 new tests).

**Checklist**

- [x] No SQL/command/XSS injection vectors in new code
- [x] Authentication/authorization enforced on all new endpoints (no new endpoints added)
- [x] No sensitive data exposed in logs, errors, or client responses
- [x] User input validated and sanitized at API boundaries
- [x] No new dependencies introduced
- [x] Error responses do not leak internal details

**Injection analysis — `milestone:` prefix handling**

The `parseInt(id.slice('milestone:'.length), 10)` call in `timelineService.ts:341` operates exclusively on values produced by the scheduling engine itself. The only strings that enter the `criticalPath` array with the `milestone:` prefix are constructed in the same service from DB-sourced integer milestone IDs (`milestone:${milestoneId}` where `milestoneId` is a `number` from `milestoneMap.keys()`). No user-controlled input reaches this code path. `parseInt` with radix 10 is the correct call and produces a safe integer.

**DB-write guard**

The `scheduled.workItemId.startsWith('milestone:')` guard in section 7 of `autoReschedule` correctly prevents virtual CPM nodes from being persisted. The filter is applied before any DB write loop, so there is no TOCTOU risk.

**`isCritical` field exposure**

The new `isCritical: boolean` field on `TimelineMilestone` is a computed boolean derived from the CPM result — it contains no PII, internal IDs, or implementation details beyond what the existing `criticalPath` work-item array already exposes. No new data-exposure risk.

**Ghost diamond isolation**

The ghost polygon always uses `strokeWidth={1.5}` and `isGhost` prop bypasses the `isCritical && !isGhost ? 3 : 2` branch correctly — no styling bleed.

PR #628, opened 2026-03-09
fix(budget): include household item budgets in budget source usage calculation

### LINKED ISSUE #626: fix(budget): Budget source usedAmount ignores household item budgets
## Problem

`computeUsedAmount` in `budgetSourceService.ts` only sums `planned_amount` from `work_item_budgets`. It ignores `household_item_budgets` that also reference budget sources, undercounting usage for any budget source that finances household items.

## Expected Behavior

`computeUsedAmount` should sum planned amounts from BOTH `work_item_budgets` AND `household_item_budgets` that reference the budget source.

## Actual Behavior

Only `work_item_budgets` are counted. Budget sources used by household items show artificially low usage.

## Reproduction Steps

1. Create a budget source
2. Create a household item budget referencing that budget source with a planned amount
3. View the budget source — `usedAmount` will be 0 instead of the household item budget's planned amount

## Acceptance Criteria

- Given a budget source referenced by both work item budgets and household item budgets
- When the budget source's used amount is computed
- Then the used amount includes the sum of `planned_amount` from both `work_item_budgets` AND `household_item_budgets`

steilerDev:
## Summary

- Fixed `computeUsedAmount` in `budgetSourceService.ts` to sum planned amounts from both `work_item_budgets` AND `household_item_budgets` via UNION ALL
- Fixed `countWorkItemReferences` (renamed to `countBudgetLineReferences`) to also check household item budgets when blocking source deletion
- Added 6 new test scenarios covering household-only, mixed, isolation, and delete-blocking cases

Fixes #626

## Test plan
- [x] 6 new unit tests added (97 total, all passing)
- [ ] CI quality gates pass


steilerDev (commented):
**[product-architect]**

This bug fix is architecturally sound. Recommending approval.

**What was verified:**

- **SQL correctness**: Both UNION ALL subqueries are valid SQLite. Using `UNION ALL` (not `UNION`) is correct since we want to sum all rows without deduplication.
- **Parameterized queries**: Both `computeUsedAmount` and `countBudgetLineReferences` use Drizzle's `sql` tagged template, which produces bind parameters. No SQL injection risk.
- **Test coverage**: 6 new test scenarios provide good coverage -- HI-only, mixed WI+HI, per-source isolation, delete blocking (HI-only, unrelated source, both types). Adequate for the scope of the fix.
- **Pattern consistency**: The switch from Drizzle query builder to `db.get<T>(sql`...`)` is the right call since Drizzle's builder does not support UNION ALL. This pattern is consistent with how the codebase handles queries that exceed the ORM's capabilities.
- **Rename**: `countWorkItemReferences` to `countBudgetLineReferences` accurately reflects the broader scope.

**Follow-up note (informational, not blocking):** `computeClaimedAmount` and `computeUnclaimedAmount` (lines 100-127) still only join through `workItemBudgets`. Since `invoice_budget_lines` has a `household_item_budget_id` FK, claimed/unclaimed amounts from household item invoices would not be counted by those functions. This is a pre-existing gap unrelated to bug #626 but worth tracking as a separate issue.

Issue #791, opened 2026-03-13
Budget Sources: Replace numeric grid with visual bar charts


steilerDev:
**As a** homeowner, **I want** the Budget Sources page to show a horizontal bar chart per source instead of a flat grid of 6-7 numeric values **so that** I can instantly assess the financial health of each funding source at a glance.

**Priority**: Should Have

### Acceptance Criteria

- [ ] **AC-1**: Given the Budget Sources page, When viewing a source card, Then a horizontal bar chart replaces the numeric amount grid, showing claimed, paid, projected min, and projected max as color-coded segments against the total budget
- [ ] **AC-2**: Given hovering (desktop) or tapping (mobile) a bar segment, When the tooltip appears, Then it displays the segment label, its value, and "Remaining: $X" relative to the total budget
- [ ] **AC-3**: Given a source with metadata (type badge, status badge, rate, terms), When viewing the card, Then this metadata is displayed below the bar chart (not mixed in with amounts)
- [ ] **AC-4**: Given both light and dark mode, When viewing the bar chart, Then segment colors use the existing BudgetBar token system (`--color-budget-claimed`, `--color-budget-paid`, etc.) and are visually consistent with the budget overview bar
- [ ] **AC-5**: Given a source where projectedAmount > totalAmount, When viewing the bar, Then overflow is visually indicated (matching the existing BudgetBar overflow pattern)

### Notes

- The existing `BudgetBar` component (`client/src/components/BudgetBar/BudgetBar.tsx`) already supports segments, overflow indication, tooltips, and multiple height sizes — reuse this component or its pattern
- Available data fields on each source: `totalAmount` (bar max), `claimedAmount`, `unclaimedAmount`, `paidAmount`, `projectedAmount`, `usedAmount`, `availableAmount`, `actualAvailableAmount`
- Bar segments should be: **Claimed** (green), **Paid/Unclaimed** (blue), **Projected Min** (light), **Projected Max** (lighter)
- Source name, type badge, status badge, terms, and rate should remain visible as metadata below the bar
- This is user feedback requesting the same visual bar chart pattern used on the Budget Overview page

PR #792, opened 2026-03-13
feat(budget-sources): replace numeric grid with visual bar charts (#791)

steilerDev:
## Summary

Replaced the `sourceAmounts` numeric grid and `plannedAllocation` line on the Budget Sources page with an interactive BudgetBar component visualization.

### Changes

- **SourceBarChart component**: Module-level sub-component rendering:
  - BudgetBar with 4 segments (claimed, paid unclaimed, projected, allocated)
  - Hover tooltip showing segment details and percentage of total
  - Bar legend with color indicators and values
  - Summary row: total, available, planned, interest rate (when applicable)

- **Segment logic** (incremental, so segments sum without overlap):
  - Claimed: `source.claimedAmount`
  - Paid (unclaimed): `Math.max(0, source.paidAmount - source.claimedAmount)`
  - Projected: `Math.max(0, source.projectedAmount - source.paidAmount)`
  - Allocated: `Math.max(0, source.usedAmount - Math.max(source.projectedAmount, source.paidAmount))`

- **Styling**: Reused `barTooltipAnchor` + `segmentTooltip` CSS pattern from BudgetOverviewPage
  - Tooltip hidden on mobile (max-width: 767px)
  - All colors from design tokens (no raw hex values)
  - Grid layout for legend: dot + label + value (right-aligned)

### Files Modified

- `client/src/pages/BudgetSourcesPage/BudgetSourcesPage.tsx`: Added SourceBarChart component, updated imports, replaced JSX
- `client/src/pages/BudgetSourcesPage/BudgetSourcesPage.module.css`: Removed old `.sourceAmounts` / `.amountGroup` / `.plannedAllocation`, added new chart classes

### Verification

- [x] No raw hex colors in CSS (all use design tokens)
- [x] BudgetBar imported with `.js` extension
- [x] BudgetBarSegment imported as `type`
- [x] useCallback added to react imports
- [x] SourceBarChart is module-level, not nested
- [x] Interest rate shown only when `interestRate != null`
- [x] Responsive: tooltip suppressed on mobile
- [x] sourceTerms block preserved below chart


steilerDev (commented):
**[ux-designer]**

Design compliance review for PR #792 — Budget Sources bar chart replacement.

---

## Summary

The implementation is well-structured and reuses the existing `BudgetBar` component correctly. Token adherence in the CSS module is strong. Two medium findings and several low/informational items are noted below.

---

## Findings

### Medium — `color-mix()` value not a CSS custom property (dark mode correctness)

**File:** `BudgetSourcesPage.tsx`, `SourceBarChart`, allocated segment definition

```tsx
color: 'color-mix(in srgb, var(--color-budget-projected) 50%, transparent)',
```

`color-mix()` is a runtime CSS computation, not a design token. The computed value is passed as an inline `style` prop, which means it bypasses the token system and cannot be overridden by the `[data-theme="dark"]` layer.

In dark mode, `--color-budget-projected` resolves to `rgba(59, 130, 246, 0.25)` — already a semi-transparent value. Mixing that 50% with transparent produces `rgba(59, 130, 246, 0.125)`, which is nearly invisible on the dark track (`--color-budget-track` = `var(--color-slate-600)`). The allocated segment will be invisible in dark mode.

**Correct approach:** Introduce a dedicated token in `tokens.css` Layer 2:

```css
/* in :root */
--color-budget-allocated: color-mix(in srgb, var(--color-budget-projected) 50%, transparent);

/* in [data-theme="dark"] */
--color-budget-allocated: rgba(59, 130, 246, 0.18);
```

Then reference it as `color: 'var(--color-budget-allocated)'` in the segment definition. This keeps the token system as the single source of truth for both modes.

---

### Medium — Tooltip missing `prefers-reduced-motion` guard

**File:** `BudgetSourcesPage.module.css`

The `.segmentTooltip` appears and disappears via conditional rendering (no CSS animation), which is acceptable. However, the `.barTooltipAnchor` wrapper is hidden at mobile with `display: none` and shown at wider viewports — that part is correct. The concern is the absence of any `@media (prefers-reduced-motion: reduce)` block in the new CSS. The `BudgetBar` segments themselves have `transition: filter var(--transition-normal)` (in `BudgetBar.module.css`), which fires on every hover. While that is in an existing file, the new code introduces additional hover state transitions (the tooltip appearing/disappearing via `useState`). No CSS animation was added here, so there is nothing to suppress in this file directly — this is informational rather than a blocker, but worth confirming: if any entrance animation is added to `.segmentTooltip` in a future iteration, a `prefers-reduced-motion` guard is mandatory.

**Severity downgrade to informational** — no CSS animation is present in this PR. No action required now.

---

### Low — Legend dot size is hardcoded, not a token

**File:** `BudgetSourcesPage.module.css`, `.barLegendDot`

```css
.barLegendDot {
  width: 8px;
  height: 8px;
  border-radius: var(--radius-circle);
  ...
}
```

`8px` is not a spacing token. The nearest tokens are `--spacing-1` (4px) and `--spacing-2` (8px). `--spacing-2` maps exactly to `0.5rem` / 8px and should be used:

```css
.barLegendDot {
  width: var(--spacing-2);
  height: var(--spacing-2);
  border-radius: var(--radius-circle);
  ...
}
```

---

### Low — `role="status"` on tooltip anchor conflicts with `aria-live="polite"`

**File:** `BudgetSourcesPage.tsx`, tooltip JSX

```tsx
<div className={styles.barTooltipAnchor} role="status" aria-live="polite">
```

`role="status"` is an implicit live region with `aria-live="polite"` and `aria-atomic="true"`. Adding explicit `aria-live="polite"` is redundant and may cause double-announcements in some screen readers. The correct pattern (matching the established `srAnnouncement` pattern used elsewhere in the codebase) is to use one or the other:

```tsx
<div className={styles.barTooltipAnchor} role="status" aria-atomic="true">
```

Remove the explicit `aria-live="polite"` and let `role="status"` carry the implicit live region semantics.

---

### Informational — Tooltip inverse-surface dark mode not explicitly guarded

**File:** `BudgetSourcesPage.module.css`, `.segmentTooltip`

```css
.segmentTooltip {
  background: var(--color-bg-inverse);
  color: var(--color-text-inverse);
  ...
}
```

This correctly mirrors the GanttTooltip pattern. In light mode, `--color-bg-inverse` = `gray-800` (dark surface) and `--color-text-inverse` = `white`. In dark mode, both flip via Layer 3 (`gray-100` / `gray-900`). The semantic tokens handle this automatically — no explicit `[data-theme="dark"]` block is needed for the tooltip surface itself, unlike GanttTooltip which has subdued color local properties that need manual overriding.

This is the correct pattern. No action required — noting it as confirmation that the implementation is consistent with the GanttTooltip precedent.

---

### Informational — Mobile breakpoint hides tooltip correctly

**File:** `BudgetSourcesPage.module.css`, responsive block

```css
.barTooltipAnchor {
  display: none;
}
```

This is scoped inside the mobile media query block and correctly suppresses the hover tooltip on touch/small viewports. Pattern is consistent with the spec intent (tooltip is a pointer-device enhancement). Confirmed correct.

---

### Informational — `summaryDivider` uses `--color-border-strong`

**File:** `BudgetSourcesPage.module.css`

```css
.summaryDivider {
  color: var(--color-border-strong);
  ...
}
```

`--color-border-strong` is `gray-300` in light mode and `slate-400` in dark mode. For a text pipe character (|), using a border token as a text color is a minor semantic mismatch. The visual result is acceptable — it renders as a muted separator — but `--color-text-muted` (`gray-500` / `slate-300`) would be more semantically accurate and produce a slightly higher-contrast separator. This is a style preference, not a correctness issue.

---

## Token Adherence Summary

All spacing, typography, shadow, z-index, and border-radius values use design tokens correctly. The budget bar tokens (`--color-budget-claimed`, `--color-budget-paid`, `--color-budget-projected`, `--color-budget-overflow`) are all referenced correctly. The inverse-surface tooltip pattern (`--color-bg-inverse` / `--color-text-inverse`) correctly mirrors GanttTooltip. The mobile `display: none` on the tooltip anchor is correct. The `aria-hidden="true"` on legend dots is correctly applied.

The only values that escape the token system are the `8px` legend dot size (Low) and the `color-mix()` allocated segment color (Medium).

---

## Verdict

**Request changes** — the `color-mix()` allocated segment will render nearly invisible in dark mode. This is a functional dark mode correctness issue. The legend dot size fix is a straightforward token substitution. Both are small, targeted edits.

Note: As a reviewer on this PR I cannot use `--request-changes` via `gh pr review` — posting as `--comment` instead. The verdict above reflects what would have been a request-changes review.

PR #953, opened 2026-03-17
fix(i18n): fix delete button translation bugs and vendor name key

### LINKED ISSUE #952: fix(i18n): fix delete button translation bugs
**As a** user, **I want** delete buttons and headers to display correct text **so that** I can safely interact with destructive actions.

**Parent Epic**: #915
**Priority**: Must Have

### Description

UAT validation revealed that several translation keys resolve to incorrect values — either returning a nested object instead of a string (causing "[object Object]" or error display), or showing raw translation keys in the UI.

### Items to Fix

- [ ] **HI detail delete button** — shows error/broken text because the translation key returns an object (nested key structure) instead of a flat string
- [ ] **Milestones detail delete button** — delete button label is missing from translation files, showing raw key or empty text
- [ ] **Vendors header** — page header displays raw key "Common.Name" instead of the translated vendor name/title

### Acceptance Criteria

- [ ] AC1: Household Item detail page delete button displays correct translated label (e.g., "Löschen" in German, "Delete" in English)
- [ ] AC2: Milestone detail delete button displays correct translated label
- [ ] AC3: Vendors page header displays the correct translated title (not "Common.Name")
- [ ] AC4: No translation key paths (e.g., "Common.Name", "HouseholdItem.Delete") are visible as raw text in the UI
- [ ] AC5: No "[object Object]" text appears anywhere in the UI from incorrect translation key resolution

### Notes

- The HI delete button bug is likely caused by a translation key like `t('HouseholdItem.Delete')` where `HouseholdItem.Delete` is a namespace/object in the JSON, not a leaf string. Fix by using the correct leaf key path.
- The Vendors header bug suggests the `t()` call is receiving a key path like `Common.Name` literally rather than resolving it.

steilerDev:
Fixes #952

### DIFF
--- a/client/src/pages/HouseholdItemDetailPage/HouseholdItemDetailPage.test.tsx
+++ b/client/src/pages/HouseholdItemDetailPage/HouseholdItemDetailPage.test.tsx
@@ -2,7 +2,7 @@
  * @jest-environment jsdom
  */
 import { jest, describe, it, expect, beforeEach } from '@jest/globals';
-import { render, screen, waitFor } from '@testing-library/react';
+import { render, screen, waitFor, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { MemoryRouter, Routes, Route, useLocation } from 'react-router-dom';
 import type * as HouseholdItemsApiTypes from '../../lib/householdItemsApi.js';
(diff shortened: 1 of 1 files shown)


steilerDev:
Superseded by #955 which includes these fixes along with comprehensive i18n completion.

PR #1076, opened 2026-03-19
fix(search-picker): prevent special option selection loop with empty id

### LINKED ISSUE #1073: Bug: Parent Area field in Manage page cannot be cleared (SearchPicker special option loop)
## Bug Report

**Parent Epic**: #915 (EPIC-17: i18n Support)

### Description

When creating a new area in the Manage page, the "Parent Area" field cannot be cleared. The field defaults to showing "No Area" as a selected state instead of showing an empty input with a placeholder. After selecting a parent area, there is no way to clear back to no parent.

### Root Cause (already identified)

In `SearchPicker.tsx`, the `selectedSpecial` check (line 65) matches the "No Area" special option (`{id: '', label: ...}`) against the initial empty value (`''`), immediately rendering the "selected special" display instead of the search input. Clicking the clear button calls `onChange('')` which keeps the value at `''`, re-matching the special option — creating an inescapable loop.

### Steps to Reproduce

1. Navigate to the Manage page
2. Click to create a new area
3. Observe the "Parent Area" field — it shows "No Area" as selected instead of an empty input
4. Try to clear the selection — the field immediately re-selects "No Area"

### Expected Behavior

- Default state: empty input field with "No Parent" placeholder text
- Clicking into the field opens the dropdown with "No Area" special option + searchable area list
- Selecting an area shows it as selected with a clear (×) button
- Clicking clear returns to the empty input state
- Selecting "No Area" explicitly from the dropdown shows it as selected with a clear button

### Actual Behavior

- Field immediately shows "No Area" as selected, blocking access to the input
- Clear button calls `onChange('')` which re-triggers the special option match
- User is trapped in "No Area" selected state

### Acceptance Criteria

- [ ] Given a new area form, when the form loads, then the Parent Area field shows an empty input with placeholder text
- [ ] Given the Parent Area field is empty, when I click/focus the input, then I see a dropdown with "No Area" option and the list of available areas
- [ ] Given I select a parent area from the dropdown, when I click the clear button, then the field returns to the empty input state
- [ ] Given I explicitly select "No Area" from the dropdown, when I see the selected state, then I can clear it with the × button to return to the empty input

### Affected Components

- `client/src/components/SearchPicker/SearchPicker.tsx` — core fix needed
- `client/src/pages/ManagePage/ManagePage.tsx` — may need placeholder text update

### Notes

The fix likely involves distinguishing between "no value selected yet" (empty/null initial state) and "user explicitly chose the special option" in the SearchPicker component. The special option match should not trigger on the default empty value — only when the user has actively selected it.

steilerDev:
## Summary

Fixes #1073 — Parent Area field in Manage page cannot be cleared when using a special option with an empty id.

## Root Cause

`SearchPicker.tsx` derived `selectedSpecial` from the `value` prop without tracking user intent:

```typescript
const selectedSpecial = specialOptions?.find((opt) => opt.id === value) ?? null;
```

When `value=''` and a special option had `id: ''` (like "No Area"), this computed value matched immediately on initial render, showing the selected display instead of the search input. Clicking clear called `onChange('')`, which re-triggered the same match — creating an inescapable loop.

## Solution

Added explicit `specialSelected` state flag to distinguish "value happens to equal special id" from "user deliberately chose a special option":

```typescript
const [specialSelected, setSpecialSelected] = useState(false);

const selectedSpecial =
  specialSelected && specialOptions
    ? (specialOptions.find((opt) => opt.id === value) ?? null)
    : null;
```

Updated handlers:
- `handleSelectSpecial()`: sets `specialSelected=true` when user clicks
- `handleClear()`: sets `specialSelected=false` when user clicks clear
- External value reset effect: also resets `specialSelected=false`

## Verification

- Initial render with `value=''` now shows search input (not selected display)
- Clicking "No Area" shows selected display with clear button
- Clicking clear returns to search input
- Normal item selection/clearing unchanged

## Changes

| File | Changes |
|------|---------|
| `client/src/components/SearchPicker/SearchPicker.tsx` | +10 lines: add `specialSelected` state, update derived `selectedSpecial` logic, track state in 3 handlers + 1 effect |


steilerDev:
Closing in favor of a clean rebase on top of PR #1077's merged fix. The specialSelected approach will be rebased cleanly.

PR #1094, opened 2026-03-20
fix(manage): randomize color picker default on page load

### LINKED ISSUE #1093: Area color picker defaults to same blue on every load
**As a** homeowner, **I want** the area creation color picker to default to a random color each time the page loads **so that** newly created areas are visually distinct without requiring me to manually change the color every time.

**Priority**: Could Have

### Acceptance Criteria

- [ ] When the Manage Areas page loads, the color picker for new area creation initializes with a random color (not the hardcoded `#3b82f6`)
- [ ] On page reload, a different random default color appears
- [ ] The random color is a valid 6-digit hex color (e.g., `#a1b2c3`)
- [ ] Previously created areas retain their assigned colors (only the default for *new* areas changes)

### Notes

- **Current behavior**: `client/src/pages/ManagePage/ManagePage.tsx` line 40 hardcodes `const DEFAULT_COLOR = '#3b82f6'`, so every new area defaults to the same blue.
- **Suggested approach**: Replace the hardcoded constant with a function that returns a random color from a curated palette or generates a random hex value. The random color should be generated once per page load (not on every render).
- This is a standalone bug fix — no parent epic.

steilerDev:
## Summary

- Introduces `generateRandomColor()` utility in `client/src/lib/colorUtils.ts` that returns a random HSL color string on each call
- Updates `ManagePage` to initialize the color picker state with a random color on every page load, so each new area/category/trade starts with a distinct default color instead of a hardcoded value
- Adds 95%+ unit test coverage for `colorUtils.ts` via `colorUtils.test.ts`

Fixes #1093

## Test plan

- [ ] Unit tests for `generateRandomColor` pass (valid HSL format, values within range, randomness verified)
- [ ] ManagePage renders with a randomly initialized color picker on each load
- [ ] CI Quality Gates pass

Co-Authored-By: Claude dev-team-lead (Sonnet) <noreply@anthropic.com>


steilerDev (commented):
**[product-architect]**

Architecture review: **No issues found.** This would be an approve if not for GitHub's self-review restriction.

**Verified:**
- `colorUtils.ts` correctly placed in `client/src/lib/` per project conventions
- HSL-constrained approach produces visually pleasant colors without needing design token changes
- `hslToHex` kept as a private helper (not exported) — good encapsulation
- `useState<string>(generateRandomColor)` correctly passes the function reference (lazy initializer), avoiding re-generation on every render
- Post-create reset calls `generateRandomColor()` (invocation) which is correct for generating a fresh color
- No shared type changes needed — color is already `string | null` in the API contract, and this change is purely UI-side default initialization
- Eliminates the `DEFAULT_COLOR` constant and `null` coalescing pattern in favor of always having a real color value — simpler state management
- Test coverage is thorough: format validation, uniqueness, deterministic boundary tests with mocked `Math.random`
- No architecture, schema, or API contract impact

No findings.

Issue #1356, opened 2026-04-25
Available Funds source rows act as per-source filter; Cost/Payback/Net per source


steilerDev:
**As a** homeowner reviewing the Cost Breakdown overview,
**I want** the Available Funds source rows themselves to act as the per-source filter (with all sources selected by default and Cost / Payback / Net columns aggregated per source),
**so that** I can see the live impact of including or excluding each fund without a parallel chip toolbar, and so that hidden sources visibly remove their cost contributions from the breakdown and Available Funds total.

### Context

This story **supersedes the chip-toolbar UX shipped in #1354 / PR #1355**. After using that implementation, the redundancy of having both source detail rows and a chip strip (and the static Available Funds total even while filtering) was confirmed to be confusing. This story converts the source rows into the filter affordance and makes Available Funds responsive to the filter.

**Parent Epic**: none (standalone enhancement; no budget epic currently open)
**Priority**: Should Have
**Supersedes**: #1354, PR #1355

#### A. Filter affordance lives on the source rows (chip strip removed)

- [ ] AC-1: When the Available Funds row is expanded, the chip strip / `BudgetSourceChip` toolbar is **not rendered**. There is no separate "all sources" clear button above the source rows.
- [ ] AC-2: Each source detail row (and the synthetic "Unassigned" row, when unassigned lines exist) is itself the toggle control. Clicking the row toggles the source's selection state.
- [ ] AC-3: The toggle control is keyboard-operable: each source row is reachable via Tab and toggles on Enter or Space. Each row exposes `aria-pressed` (or equivalent) reflecting the selected state, with an accessible name that includes the source name and selection state.
- [ ] AC-4: A screen-reader live region announces the number of currently selected sources (e.g., "3 of 4 sources selected") whenever selection changes.

#### B. Default state and visual deselection

- [ ] AC-5: On initial page load with no filter URL parameter, **all sources (including the Unassigned bucket, if present) are selected**. No source rows render in the deselected/greyed style.
- [ ] AC-6: When a source is deselected, the row renders in a greyed-out / disabled visual treatment per the UX spec (reduced opacity or muted token). The source dot color is preserved at reduced emphasis to keep the source identifiable.
- [ ] AC-7: Deselecting all sources is permitted; in that case the breakdown's Sum, Available Funds total, and all areas/items render the empty state already used by the breakdown table.

#### C. Cost / Payback / Net columns per source row

[...]

PR #466, opened 2026-03-05
feat(budget): add expandable cost breakdown table to Budget Overview

steilerDev:
## Summary

- Adds `GET /api/budget/breakdown` endpoint backed by `budgetBreakdownService` that aggregates work items, vendors, and budget lines into a 4-level cost hierarchy per budget category
- Adds `CostBreakdownTable` React component with expandable rows (category → work item → vendor → budget line), confidence labels, margin range columns, and invoice tracking
- Integrates the table into the Budget Overview page and extends `budgetOverviewApi.ts` with the new `fetchBudgetBreakdown` call; 85 tests across 3 test files cover the backend service, route integration, and frontend component

Fixes #464

## Test plan

- [ ] Unit tests pass (95%+ coverage) — `budgetBreakdownService.test.ts`, `budgetOverview.breakdown.test.ts`, `CostBreakdownTable.test.tsx`
- [ ] Integration tests pass
- [ ] Pre-commit hook quality gates pass (lint, format, typecheck, build, audit)

Co-Authored-By: Claude dev-team-lead (Sonnet 4.6) <noreply@anthropic.com>
Co-Authored-By: Claude backend-developer (Haiku 4.5) <noreply@anthropic.com>
Co-Authored-By: Claude frontend-developer (Haiku 4.5) <noreply@anthropic.com>
Co-Authored-By: Claude qa-integration-tester (Sonnet 4.5) <noreply@anthropic.com>

PR #414, opened 2026-03-03
feat(invoices): add invoice linking for household item budget lines (#413)

steilerDev:
## Summary

- Add `household_item_budget_id` FK to invoices table (migration 0011) with mutual exclusivity validation — an invoice links to either a work item budget line or a household item budget line, never both
- Update budget overview and household item budget summary to include actual invoice totals (replaces hardcoded `totalActual: 0`)
- Add household item picker with budget line dropdown to invoice create/edit modals with mutual exclusivity UX
- Display linked invoices per budget line on household item detail page

Fixes #413

## Test plan
- [ ] Unit tests pass (95%+ coverage)
- [ ] Integration tests pass
- [ ] Pre-commit hook quality gates pass

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>
Co-Authored-By: Claude backend-developer (Haiku 4.5) <noreply@anthropic.com>
Co-Authored-By: Claude frontend-developer (Haiku 4.5) <noreply@anthropic.com>
Co-Authored-By: Claude qa-integration-tester (Haiku 4.5) <noreply@anthropic.com>

Issue #29, opened 2026-02-08
2.5: Responsive Layout & Cross-Browser Compatibility


steilerDev:
## User Story

**As a** user, **I want** the application to work well on my tablet and phone as well as my desktop, **so that** I can check project progress from the construction site.

## Parent Epic

EPIC-02: Application Shell & Infrastructure (#2)

## Priority

Must Have

## Dependencies

- Story 2.4: Client-Side Routing & Application Shell Layout (#27)

## Acceptance Criteria

- [ ] 1. The application shell layout adapts to three breakpoints: mobile (<768px), tablet (768-1024px), desktop (>1024px)
- [ ] 2. On mobile, the sidebar navigation collapses into a hamburger menu
- [ ] 3. On tablet, the sidebar can be toggled open/closed
- [ ] 4. On desktop, the sidebar is always visible
- [ ] 5. Touch targets (buttons, links) are at least 44x44px on mobile/tablet
- [ ] 6. Text remains readable without horizontal scrolling on all breakpoints
- [ ] 7. The app renders correctly in the latest versions of Chrome, Firefox, Safari, and Edge

## Notes

- Requirements Section 3.2 specifies responsive design with desktop and tablet support, mobile-friendly
- Requirements Section 3.4 specifies touch-friendly interface on mobile devices
- CSS Modules are used for styling -- breakpoint mixins or CSS custom properties should establish the responsive foundation
- This story builds on the shell layout from Story 2.4 and makes it responsive

steilerDev:
**[uat-validator]**

## UAT Scenarios: Responsive Layout & Cross-Browser Compatibility (Issue #29)

### Preconditions
- Application is built and running
- Story #27 (Client-Side Routing & Application Shell Layout) is complete
- Testing browsers installed: Chrome, Firefox, Safari (macOS/iOS), Edge
- Browser DevTools or responsive design mode available for viewport resizing

---

### Scenario 1: Desktop Layout Shows Persistent Sidebar
**Acceptance Criteria**: 1, 4

- **Given**: User opens the application in a desktop browser
- **When**: User resizes the browser window to 1280px width (desktop breakpoint >1024px)
- **Then**: 
  - The sidebar is fully visible on the left side of the screen
  - The sidebar does NOT have a close button or toggle icon
  - The main content area occupies the remaining horizontal space
  - Navigation links are fully visible with labels
- **Verification Method**: Manual

---

[...]

PR #1361, opened 2026-04-26
feat(budget): move per-source filter to server-side via ?deselectedSources=

### LINKED ISSUE #1360: Move per-source filter to server-side via ?deselectedSources= so subsidies recompute against the filtered cost set
**As a** homeowner using the Budget Overview's per-source filter,
**I want** the server to recompute the entire breakdown (including subsidy oversubscription, area/item totals, and per-section sums) against only the lines for my currently selected sources,
**so that** the numbers I see on screen are internally consistent — Cost, Payback, Net, oversubscription excess, and the Sum row all describe the same filtered project, not a mix of "filtered cost minus project-wide subsidies".

**Parent Epic**: none (cross-cutting refactor on top of #1356 / #1358)
**Priority**: Should Have
**Supersedes architect decision from #1356**: the prior decision that "subsidy adjustments stay project-wide while filter is applied" is reversed by this story. Subsidy oversubscription is now recomputed against the filtered cost set.

---

## Problem

Issues #1356 / #1358 (just merged) ship a per-source filter UX where deselecting a source in the Available Funds rows hides its lines and aggregates the visible totals client-side. While the cost figures shrink as expected, the subsidy oversubscription excess stays project-wide. This produces nonsensical math — e.g. when filtering to a single small source, the "Payback" can exceed the visible "Cost" because the oversubscription excess was computed against the un-filtered project.

User feedback that drove this:

> Subsidies oversubscription needs to be adjusted, if the input is filtered — since it's staying constant it's adding additional cost exceeding the subsidy itself. I'm not sure what the best way to handle this would be because if the subsidy is applied to multiple sources and is oversubscribed, how would we handle calculation? Applying oversubscription based on the assumption that only the visible items are calculated.

The root cause is that filtering happens client-side after the breakdown has already been aggregated and the subsidy engine has already run. A correct filtered view requires re-running the entire pipeline against the filtered line set on the server.

---

## Solution

Move filtering server-side. Extend `GET /api/budget/breakdown` to accept a `?deselectedSources=<id1>,<id2>,unassigned` query parameter. The breakdown service applies this filter at the very top of its pipeline (before any aggregation, before the subsidy engine runs), so every emitted aggregate — item totals, area subtotals, section totals, subsidy adjustments, per-source projected ranges — describes a self-consistent filtered project. The client refetches the breakdown on every chip toggle and removes its in-memory aggregation logic.

---

### Server (`GET /api/budget/breakdown`)

1. **Given** a request to `GET /api/budget/breakdown` with no query params, **when** processed, **then** the response is identical to the pre-refactor response (full project-wide breakdown — backward compatible).
2. **Given** a request with `?deselectedSources=` (empty string) or `?deselectedSources` (no value), **when** processed, **then** the response is identical to the no-param case.
3. **Given** a request with `?deselectedSources=<id1>,<id2>`, **when** processed, **then** lines whose `budget_source_id` is in the comma-separated list are excluded from the entire pipeline (queries A and C above) before any aggregation runs.
4. **Given** a request with `?deselectedSources=unassigned` (literal string), **when** processed, **then** lines with `budget_source_id IS NULL` are excluded.
5. **Given** a request with `?deselectedSources=<unknown-uuid>,unassigned,<real-id>`, **when** processed, **then** the unknown UUID is silently ignored (no 400) and only `unassigned` + `<real-id>` are filtered. _Reason: URL state may reference a source the user has since deleted; we must degrade gracefully rather than break the page._
6. **Given** any combination of deselected sources, **when** the breakdown is computed, **then** all `BreakdownArea.projectedMin/Max/actualCost/subsidyPayback/rawProjectedMin/Max/minSubsidyPayback`, all `BreakdownWorkItem`/`BreakdownHouseholdItem` totals, and the section-level `BreakdownTotals` reflect only the surviving (non-filtered) lines.
7. **Given** any combination of deselected sources, **when** the subsidy calculation engine runs, **then** it runs against the filtered set of budget lines only — `subsidyAdjustments[].minExcess`, `subsidyAdjustments[].maxExcess`, `subsidyAdjustments[].maxPayout` reflect the over-/under-subscription state of the filtered project. A subsidy that no longer matches any surviving line MUST NOT appear in `subsidyAdjustments` (no zero-excess rows).
8. **Given** an item or area whose lines are entirely filtered out, **when** the breakdown is built, **then** the item/area is NOT emitted in the response (server-side pruning, not client-side hiding). The pruning rules already in `buildAreaBreakdown` apply; the only change is that the input set is pre-filtered.
9. **Given** a filtered request, **when** building the `budgetSources[]` array, **then** ALL configured sources still appear (the array is the master list of sources for chip rendering). Their `projectedMin/Max` reflects only surviving lines for each source. **See Architect Question A** for the exact shape of deselected-source entries.
10. **Given** a filtered request, **when** the engine produces per-subsidy payback, **then** the `BudgetSourceSummaryBreakdown` for each still-selected source contains a per-source attributed `subsidyPayback` (min and max). _**See Architect Question B.** Without this, the client cannot render the per-source "Payback" column without re-implementing the engine — which is the duplication this story exists to eliminate._

### Client (`BudgetOverviewPage` + `CostBreakdownTable`)

11. **Given** the user toggles a chip in the Available Funds source row, **when** the URL state updates, **then** `fetchBudgetBreakdown` is re-invoked with the new `deselectedSources` query value and the table re-renders against the server's response.
12. **Given** a refetch is in flight, **when** the table is re-rendered, **then** the previous breakdown remains visible (stale-while-revalidate). The table MUST NOT flicker to an empty state. A subtle loading indicator (e.g. opacity dim, spinner badge) is acceptable; full skeleton replacement is not.
13. **Given** the user toggles multiple chips in rapid succession (≤ 200ms between toggles), **when** requests would otherwise pile up, **then** requests are coalesced via debounce (**See Architect Question D for the exact debounce window**) so only the final state is fetched.
14. **Given** a refetch fails (network error / 500), **when** the error surfaces, **then** the previous breakdown remains visible and an inline error toast/banner explains the failure. The chip selection state remains as the user set it.
15. **Given** the URL on page load contains `?deselectedSources=…`, **when** the page mounts, **then** the initial `fetchBudgetBreakdown` call sends the query param so the first render is already filtered (no flash of unfiltered content).
16. **Given** any deselected-source state, **when** the table renders, **then** all aggregate values (Cost, Payback, Net, Sum row, every area subtotal at every depth, every item row, the WI/HI section headers, and the Subsidy Adjustments rows) come directly from the server response. _No client-side aggregation across multiple lines remains._
17. **Given** the URL is `?deselectedSources=` (none deselected), **when** the table renders, **then** every chip is rendered as selected and the breakdown matches the un-filtered project — same as a fresh page load.

### Code deletion (the refactor's main payoff)

The following client-side code MUST be deleted. The story is not Done until each is removed:

18. `computeFilteredAggregates` function (`CostBreakdownTable.tsx` ~lines 172–287) — DELETE.
19. `computePerSourcePayback` function (`CostBreakdownTable.tsx` ~lines 108–156) — DELETE. Per-source payback now comes from the server (AC #10).
20. `filteredAdjustedTotalPayback` `useMemo` (`CostBreakdownTable.tsx` ~lines 1199–1214) — DELETE.
21. `filteredRawProjected` `useMemo` (`CostBreakdownTable.tsx` ~lines 1163–1192) — DELETE.
22. `filteredAvailableFunds` `useMemo` (`CostBreakdownTable.tsx` ~lines 1124–1129) — DELETE. _(`overview.availableFunds` may also need a per-source-aware variant on the server; however that lives in the `/budget/overview` endpoint, not `/budget/breakdown`. Out of scope for this story — see Open Questions.)_
23. `unassignedAllocatedCost` `useMemo` (`CostBreakdownTable.tsx` ~lines 1132–1159) — DELETE.
24. `visibleLineIds` `useMemo` (`CostBreakdownTable.tsx` ~lines 1086–1115) — DELETE.
25. `BreakdownContext.filteredAggregates` field, `BreakdownContext.visibleLineIds` field, `FilteredAggregates` interface, `FilteredEntityTotals` interface — DELETE all four (lines 36–42, 158–170).
26. `areaHasVisibleLines` helper (`CostBreakdownTable.tsx` ~lines 388–401) — DELETE. Server prunes empty areas (AC #8).
27. All `if (hasSourceFilter && …) return null` cascade-hide guards in `BudgetLineRow`, `WorkItemRow`, `HouseholdItemRow`, `WorkItemAreaSection`, `HouseholdItemAreaSection` — DELETE. The server response already excludes filtered items.
28. All `filteredAggregates`-conditional ternaries in row render code (`WorkItemRow`, `HouseholdItemRow`, `WorkItemAreaSection`, `HouseholdItemAreaSection`, and the WI/HI section header rows in `CostBreakdownTable`) — DELETE in favour of unconditional reads from the server-provided `area.*` / `item.*` fields.

### Tests

29. Backend service tests in `budgetBreakdownService.test.ts` cover: empty/missing param (AC #1, #2), single-source filter (AC #3), `unassigned` filter (AC #4), unknown-id graceful behaviour (AC #5), subsidy re-run produces correct excess for filtered set (AC #7), pruning of items/areas with no surviving lines (AC #8), and `budgetSources[]` shape under filter (AC #9, #10).
30. Backend route test in `budgetOverview.breakdown.test.ts` covers: query-param parsing (single value, comma list, empty, missing, with `unassigned`), 401 for unauthed.
31. Client integration test covers: chip toggle triggers refetch with correct query param, stale-while-revalidate behaviour during refetch (AC #12), debounce coalesces rapid toggles (AC #13), error path keeps previous data (AC #14), URL-on-mount triggers filtered first fetch (AC #15).
32. E2E test (`e2e/tests/budget/cost-breakdown-source-filter.spec.ts`): user toggles a chip, sees Cost/Payback/Net/Sum all change consistently and Subsidy Adjustments excess updates to match the filtered subsidy run. The "filtered cost lower than visible payback" bug case from the user feedback is the canonical test scenario.

---

### Question A — `BudgetSourceSummaryBreakdown` shape under filter

When a source is deselected, what should its entry in `budgetSources[]` look like?

- **(a)** Omit the entry entirely. _Client must keep its own list of all sources from `fetchBudgetSources` to render chips._
- **(b)** Include the entry with `totalAmount` from source config but `projectedMin/Max = 0` (since its lines are filtered out and therefore not part of the visible cost picture).
- **(c)** Include the entry with both `totalAmount` AND `projectedMin/Max` reflecting the **un-filtered** values for that source (so the chip still shows what it would contribute if re-selected, helping the user reason about "what am I filtering away").

**PO recommendation: (c)** — keeps the chip informative when deselected. The mental model is "the chip shows you what's available; the table shows you what you've selected". But we need the architect to confirm this doesn't muddy the "filtered" semantics elsewhere.

### Question B — per-source `subsidyPayback` attribution in the response

The current `BudgetSourceSummaryBreakdown` exposes `projectedMin/Max` only — no payback. The client today computes per-source payback via `computePerSourcePayback` (pro-rata weighting by line cost). Once filtering moves server-side, the server has all the inputs to compute this directly from a single engine run against the filtered line set, with the same pro-rata attribution rule.

**Question:** Should the response add `subsidyPaybackMin` / `subsidyPaybackMax` fields to `BudgetSourceSummaryBreakdown` (or a parallel array keyed by source ID)?

**PO recommendation: yes.** Required for AC #10 / #19 — without it the client can't render per-source Payback without duplicating the engine. The architect should specify the exact field names and whether `unassigned` should appear as a synthetic entry in this map.

### Question C — performance

A network round-trip per chip click (~50–200ms on SQLite) is the agreed tradeoff. Confirm acceptance and confirm no caching layer is needed for v1.

**PO recommendation: accepted** — the user explicitly accepted this tradeoff in the conversation that drove this story. Mobile users may see a transient loading state, which AC #12 addresses.

### Question D — request coalescing / debounce

When the user toggles multiple chips in quick succession, do we (i) fire a request per toggle and let the latest response win, (ii) debounce by N ms before firing, or (iii) cancel in-flight requests on each new toggle?

**PO recommendation: debounce 50ms** — short enough to feel instant, long enough to coalesce double-clicks and "select all" sequences. The architect should specify the exact value and whether `AbortController` should also cancel superseded in-flight requests.

---

## Out of Scope

- No new UI design surface. Chip rendering, filter affordances, and table layout are unchanged. **No `ux-designer` spec needed.**
- No backend schema changes. **No migration.** The filter is a query parameter only.
- No changes to source CRUD (`/api/budget/sources`).
- The `availableFunds` figure on the `BudgetOverview` summary is computed by `/api/budget/overview`, not `/api/budget/breakdown`. Whether to filter that endpoint as well is a follow-up. For this story, AC #22's deletion of `filteredAvailableFunds` is replaced by a server-side `availableFunds` value scoped to selected sources — but that requires extending `/api/budget/overview`. **Architect to decide:** include `/api/budget/overview` filter in this story, or split into a follow-up. PO leans toward splitting (this story is already large).
- No changes to the print/export view.

---

## Notes / Dependencies

- **Supersedes** the "subsidy adjustments stay project-wide" decision from #1356.
- Builds on the URL state plumbing already merged in #1356 (`BudgetOverviewPage.tsx` lines 184–223). No URL semantics change.
- `BudgetBreakdown` shape stays the same except for the additions decided by Question B.
- Affected files:
  - `server/src/routes/budgetOverview.ts` (route handler — accept query param)
  - `server/src/services/budgetBreakdownService.ts` (filter at top of pipeline; re-run subsidy engine on filtered lines; prune)
  - `server/src/services/budgetBreakdownService.test.ts` (new tests per AC #29)
  - `server/src/routes/budgetOverview.breakdown.test.ts` (new tests per AC #30)
  - `shared/src/types/budgetBreakdown.ts` (potential additions per Question B)
  - `client/src/lib/budgetOverviewApi.ts` (extend `fetchBudgetBreakdown` to accept deselected source IDs)
  - `client/src/pages/BudgetOverviewPage/BudgetOverviewPage.tsx` (refetch on chip toggle; debounce; stale-while-revalidate)
  - `client/src/components/CostBreakdownTable/CostBreakdownTable.tsx` (massive deletion — see ACs #18–28)
  - `client/src/components/CostBreakdownTable/CostBreakdownTable.test.tsx` (delete tests for deleted code; add tests for new server-driven render paths)
  - `e2e/tests/budget/` (new E2E spec per AC #32)

steilerDev:
## Summary

Moves the per-source filter from the client to the server, fixing the subsidy
oversubscription consistency bug introduced by #1356/#1358 (filtered Cost
could be lower than displayed Payback because the subsidy engine ran against
the unfiltered project).

- `GET /api/budget/breakdown?deselectedSources=<id1>,<id2>,unassigned` accepts
  a comma-separated list of deselected source IDs (and the `unassigned`
  literal). Empty/missing param = full project-wide breakdown.
- Backend filters lines at the top of the pipeline, re-runs the subsidy engine
  against the filtered cost set, and emits filter-aware aggregates +
  per-source pro-rata payback (`subsidyPaybackMin`, `subsidyPaybackMax`).
- Items / areas with no surviving lines are pruned server-side.
- Source detail rows still show their unfiltered Cost (architect decision A:
  the deselected source stays informative — user sees what they're filtering
  away).
- Client refetches on each chip toggle with 50ms debounce + AbortController +
  stale-while-revalidate (previous breakdown stays visible at reduced opacity).
- Removes ~530 lines of client-side aggregation (`computeFilteredAggregates`,
  `computePerSourcePayback`, all filtered memos, all cascade-hide guards).

Closes #1360

This explicitly supersedes the #1356 decision that "subsidy adjustments stay
project-wide".

## Test plan

- [ ] Quality Gates green (lint, typecheck, unit + integration tests)
- [ ] E2E Tests across 16 shards × 3 viewports green (subsidy oversubscription
  consistency scenario added — canonical bug verification)
- [ ] Backend integration tests: filter param parsing, empty/missing/unknown
  source IDs, `unassigned` literal, cascade pruning, per-source unfiltered
  projections, per-source pro-rata payback
- [ ] Frontend unit tests: debounce + AbortController + stale-while-revalidate
- [ ] API client query-string encoding


steilerDev:
**[product-architect]** Approve (verdict: approve — comment instead of `--approve` since I co-authored).

Architect decisions A/B/C/D verified end-to-end:

- **Pipeline order correct.** `filteredWiLineRows`/`filteredHiLineRows` (`budgetBreakdownService.ts:163-164`) computed at top of pipeline, before entity aggregation (line 533+) and `computeEntitySubsidyPayback` (lines 600, 764). Subsidy engine runs against the filtered set — the canonical bug from #1356/#1358 is fixed at the source.
- **Decision A (per-source `projectedMin/Max` UNFILTERED)** honored: per-source projection accumulators iterate `workItemLineRows`/`hiLineRows` (unfiltered) and reuse `wiLineInvoiceMap`/`hiLineInvoiceMap` for actuals — no leak from the filtered path.
- **Decision B (per-source `subsidyPaybackMin/Max` from FILTERED engine)** honored: `sourcePaybackMap` walks `wiEntityData`/`hiEntityData` (filtered) with pro-rata weighting by max-perspective resolved cost, falling back to equal split when total cost is 0 — sound. Deselected sources naturally yield `{min: 0, max: 0}`.
- **Cascade pruning is server-side**: items/areas with no surviving lines are dropped at aggregation, not at row render. Client deletes `computeFilteredAggregates`, `computePerSourcePayback`, all cascade-hide guards (~530 lines).
- **Empty-state UX fix is correctly scoped**: `if (!hasData && !hasSources)` (`CostBreakdownTable.tsx:1958`) only short-circuits when both areas are empty AND `budgetSources` is empty — preserves the "no data at all" path while keeping source rows visible when all sources are deselected.
- **Wiki API contract** (`wiki@771c1df`) fully documents the query param, the new `BudgetSourceSummaryBreakdown` shape with both new fields, the unfiltered/filtered split, and the "all sources always included + synthetic unassigned" behavior.
- **`discretionary-system` seeded source** flows through the standard `budgetSourceId` path with no special-case handling needed.

Nit (informational, not blocking): post-migration 0023, no production line has `budget_source_id IS NULL`, so the `'unassigned'` literal in `isLineFiltered` is effectively dead in current data. Harmless and forward-compatible — keep as defensive code.

LGTM.

steilerDev (commented):
**[security-engineer]** PR #1361 security review — APPROVED

**No blocking findings.**

Verified the three security-relevant surfaces:

**Query param parsing** (): Raw string split on `,`, trimmed, empty strings filtered out via `filter(Boolean)`. Result is a `Set<string>` consumed only by `Set.has()` lookups — no user input reaches any SQL query. Unknown IDs are silently ignored, which is the correct behavior (no error oracle). No injection vector.

**Filter implementation** (): Filtering is entirely in-memory on already-fetched row arrays (`filteredWiLineRows`, `filteredHiLineRows`). The `isLineFiltered()` helper uses only `Set.has()` and strict null equality — no SQL interpolation of user input at any point. Per-source projections iterate the *unfiltered* original rows (architect decision A), which is correct and expected.

**Authorization**: The existing `preValidation` session hook remains in place and unchanged. The new query-param path is inside the same authenticated handler. No bypass introduced.

**Frontend**: `encodeURIComponent(join(','))` correctly URL-encodes the deselected IDs before they reach the server. The `unassigned` literal is a hardcoded string in the client and is handled as a known sentinel on the server — no reflection of arbitrary user data into rendered output. No `dangerouslySetInnerHTML` or raw DOM injection found in the changed files.

**Informational — no schema validation on `deselectedSources`**: The query param has no Fastify schema constraint (no maxLength on individual IDs, no maxItems on the comma-separated list). In practice the Set is only ever used for `Set.has()` comparisons so there is no exploit path, but adding a schema-level bound (e.g., max param string length of 4096 chars) would be a defense-in-depth improvement. Low priority, consistent with prior findings on this route (#466).

Checklist: injection ✓ | auth ✓ | sensitive data ✓ | input validation ✓ | dependencies ✓ | secrets ✓

steilerDev (commented):
**[security-engineer]** PR #1361 security review -- APPROVED

**No blocking findings.**

Verified the three security-relevant surfaces:

**Query param parsing** (`budgetOverview.ts` lines 30-71): Raw string split on `,`, trimmed, empty strings filtered out via `filter(Boolean)`. Result is a `Set<string>` consumed only by `Set.has()` lookups -- no user input reaches any SQL query. Unknown IDs are silently ignored, which is the correct behavior (no error oracle). No injection vector.

**Filter implementation** (`budgetBreakdownService.ts`): Filtering is entirely in-memory on already-fetched row arrays (`filteredWiLineRows`, `filteredHiLineRows`). The `isLineFiltered()` helper uses only `Set.has()` and strict null equality -- no SQL interpolation of user input at any point. Per-source projections iterate the *unfiltered* original rows (architect decision A), which is correct and expected.

**Authorization**: The existing `preValidation` session hook remains in place and unchanged. The new query-param path is inside the same authenticated handler. No bypass introduced.

**Frontend**: `encodeURIComponent(join(","))` correctly URL-encodes the deselected IDs before they reach the server. The `unassigned` literal is a hardcoded string in the client and handled as a known sentinel on the server -- no reflection of arbitrary user data into rendered output. No `dangerouslySetInnerHTML` or raw DOM injection found in the changed files.

**Informational -- no schema validation on `deselectedSources`**: The query param has no Fastify schema constraint (no maxLength on individual IDs, no maxItems on the comma-separated list). In practice the Set is only ever used for `Set.has()` comparisons so there is no exploit path, but a schema-level bound (e.g., max param string length of 4096 chars) would be a defense-in-depth improvement. Low priority, consistent with prior findings on this route (#466).

Checklist: injection OK | auth OK | sensitive data OK | input validation OK | dependencies OK | secrets OK

steilerDev (commented):
**[product-owner]** PR #1361 review — APPROVED (advisory; cannot self-approve as repo owner)

All 31 acceptance criteria from issue #1360 verified. Architect decisions A/B/C/D honored. The canonical bug (subsidy oversubscription consistency) is fixed.

**Server (AC #1–#10)** — `getBudgetBreakdown(db, deselectedSources)` filters at the top of the pipeline (`filteredWiLineRows`/`filteredHiLineRows`). Subsidy engine re-runs on the filtered set, items/areas with no surviving lines are pruned, route accepts `?deselectedSources=` with comma-separated UUIDs and the `unassigned` literal. Unknown UUIDs degrade silently. Route defaults to empty Set (backward compatible). Backend tests Scenarios 1–10 in `budgetBreakdownService.test.ts` and Scenarios 11–15 in `budgetOverview.breakdown.test.ts` cover every server-side AC including the 401 unauth case.

**Decision A honored**: `sourceProjectedMap` iterates the UNFILTERED `workItemLineRows`/`hiLineRows`, so deselected sources keep their full projected contribution in `budgetSources[]`. `BudgetSourceSummaryBreakdown` in `shared/src/types/budgetBreakdown.ts` now carries `subsidyPaybackMin`/`subsidyPaybackMax` (Decision B), pro-rata attributed from the filtered engine run via `sourcePaybackMap`. Synthetic `unassigned` entry emitted whenever any null-source lines exist.

**Client (AC #11–#17)** — `BudgetOverviewPage.tsx` debounce + AbortController + stale-while-revalidate effect implements Decision D with `DEBOUNCE_MS = 50`. The wrapper div toggles `styles.breakdownRefetching` (opacity 0.6, pointer-events: none) so the previous breakdown stays visible during refetch — no flicker. Refetch errors render an inline `role=\"alert\"` banner with a Dismiss button; the previous breakdown is preserved (AC #14). Initial load reads `deselectedSourceIds` from the URL before the first fetch (AC #15).

**Code deletion (AC #18–#28)** — every named symbol from issue #1360 is gone: `computeFilteredAggregates`, `computePerSourcePayback`, `filteredAdjustedTotalPayback`, `filteredRawProjected`, `filteredAvailableFunds`, `unassignedAllocatedCost`, `visibleLineIds`, `FilteredAggregates`/`FilteredEntityTotals`, `areaHasVisibleLines`, all `hasSourceFilter && …` cascade guards, all `filteredAggregates`-conditional ternaries. Section/area/item rows read directly from server-provided `*.rawProjectedMin/Max` and `*.subsidyPayback`. Per-source row Payback reads `source.subsidyPaybackMin/Max`.

**Tests (AC #29–#32)** — service + route tests cover every server AC. Client tests cover debounce/abort/SWR/URL-on-mount/error path (Scenarios 24–29). E2E `budget-source-filter.spec.ts` adds the canonical subsidy oversubscription scenario (Scenario D), stale-while-revalidate (Scenario E), URL-on-mount (Scenario F), and rapid debounce coalescence (Scenario H).

**Late-cycle empty-state UX fix** — verified in `CostBreakdownTable.tsx`: early-return empty state now fires only when `!hasData && !hasSources`. When all sources are deselected (server prunes areas, but `budgetSources[]` is non-empty), the full table renders — source rows, Available Funds, Sum, Remaining Budget all visible with zeroed totals. The user is no longer trapped in an empty-state dead end. Regression test \"Scenario 24\" in `CostBreakdownTable.test.tsx` locks this in.

Security-engineer approved separately. Cleared for merge.

steilerDev (commented):
**[ux-designer]**

Reviewed the UX surfaces in PR #1361. Two findings — one critical (dark mode contrast), one medium (recurring focus token pattern). This would be `--request-changes` on an external PR; posting as a comment since this is our own PR.

---

### Critical — Dark mode contrast failure in `.breakdownErrorBanner`

The error banner uses `background: var(--color-bg-inverse)` but `color: var(--color-text-primary)` for both the banner text and the dismiss button.

In **dark mode** these tokens resolve to:

- `--color-bg-inverse` → `--color-gray-100` = `#f3f4f6` (light gray surface)
- `--color-text-primary` → `--color-slate-50` = `#f1f5f9` (near-white text)

Near-white text on a light gray background is essentially invisible — far below WCAG AA 4.5:1. The same token conflict exists on the dismiss `<button>` inside the banner.

**Fix:** On any element using `--color-bg-inverse` as its background, pair it with `--color-text-inverse` for text (resolves to `#fff` in light mode, `--color-gray-900` in dark mode — correct in both).

```css
.breakdownErrorBanner {
  background: var(--color-bg-inverse);
  color: var(--color-text-inverse);   /* not --color-text-primary */
}
.breakdownErrorBanner button {
  color: var(--color-text-inverse);   /* not --color-text-primary */
}
```

This matches the established GanttTooltip pattern documented in the Style Guide.

---

### Medium — Focus ring uses `outline` instead of `--shadow-focus` token

```css
/* current */
.breakdownErrorBanner button:focus-visible {
  outline: 2px solid var(--color-primary);
  outline-offset: 2px;
}
```

Per the design system, focus rings must use `box-shadow: var(--shadow-focus)`. `outline: 2px solid var(--color-primary)` is a recurring pattern flagged in PRs #402 and #414.

**Fix:**
```css
.breakdownErrorBanner button:focus-visible {
  box-shadow: var(--shadow-focus);
  outline: none;
}
```

---

### Approved patterns

- `.breakdownRefetching` — `opacity: 0.6`, `pointer-events: none`, `transition: opacity var(--transition-fast)` — all tokens, correctly guarded by `@media (prefers-reduced-motion: reduce) { transition: none; }`. No issues.
- New i18n keys `refetchError` / `dismissError` use `t()` correctly in both EN and DE.
- Empty-state UX fix (deselecting all sources keeps table visible) — correct behavior, no styling concerns.
- No new design tokens introduced — confirmed.

Issue #1411, opened 2026-05-12
Invoice status summary breakdown is not deposit-aware (InvoicesPage header + Dashboard pipeline card)


steilerDev:
## Context

The invoice deposits feature trilogy just shipped (#1403, #1404, #1405). #1405 added deposit-aware budget rollups across all budget paths (budget overview, budget sources, work/household-item budget summaries). However, the `InvoiceStatusBreakdown` returned by `listAllInvoices` was NOT migrated to the deposit-aware split, despite #1405's spec calling for it explicitly.

The breakdown is consumed by two prominent surfaces — both are "budget context" views per the user's mental model:

1. **InvoicesPage** (`/budget/invoices`) header — four summary cards: Pending / Paid / Claimed / Quotation, each with a total amount and count.
2. **Dashboard's `InvoicePipelineCard`** — pending + quotation totals at the bottom.

Both render `summary.<status>.totalAmount` from `InvoiceStatusBreakdown`, which is computed in `server/src/services/invoiceService.ts:271-281` by a simple `GROUP BY invoices.status, SUM(invoices.amount)` — no deposit awareness.

## Problem

After adding a `pending` deposit of €200 to an invoice with `status='quotation'` and `amount=€1000`:

- **Expected**: Quotation card shows €800 (the residual = 1000 − 200), Pending card shows €200 (the deposit).
- **Actual**: Quotation card shows €1000 (the full parent amount), Pending card shows €0.

## Expected behavior

The summary cards must reflect the same proportional split already implemented for the budget rollups:

- For each deposit `d` with status `s`, `summary[s].totalAmount` accrues `d.amount`.
- For each invoice `I` with parent status `S`, `summary[S].totalAmount` accrues the residual `I.amount − Σ d.amount` (the "Final payment" portion).
- For each deposit, `summary[s].count` increments by **0** (we don't double-count rows — only the parent invoice contributes to `count`).
- Each invoice contributes **1** to `summary[S].count` (the parent status's count), even when its residual is 0.

**Invariant:** `Σ summary[s].totalAmount` across all statuses must still equal `Σ I.amount` across all invoices, exactly (modulo currency rounding). No double counting.

## Reproduction steps

1. Create a vendor.
2. Create an invoice for vendor with `amount=1000`, `status='quotation'`.
3. Add a deposit to that invoice with `amount=200`, `status='pending'`.
4. Open `/budget/invoices` (the InvoicesPage).
5. Observe the summary header.

## Acceptance criteria

[...]

PR #1407, opened 2026-05-11
feat(invoice,budget): invoice deposits UI + deposit-aware budget rollups (#1404, #1405)

steilerDev:
## Summary

**#1404 — Invoice deposits detail-page UI**
- Adds a Deposits section to the invoice detail page with add/edit/delete controls, per-deposit state toggles (unclaimed → claimed → paid), and a read-only "Final payment" row showing the residual amount (invoice total minus sum of deposits).
- Uses shared `Modal`, `Badge`, `FormError`, and `EmptyState` components. Responsive: table layout on desktop/tablet (claimed-date column hidden below 1024 px), card list on mobile. Overflow menu implements the WAI-ARIA Menu Button pattern (ArrowUp/Down/Home/End/Escape keyboard navigation). New i18n keys under `invoiceDetail.deposits.*` in EN and DE; glossary updated: Deposit → Abschlagszahlung, Final payment → Schlusszahlung.

**#1405 — Deposit-aware budget rollups**
- Budget rollups now proportionally split each invoice budget-line contribution between its deposits (each under the deposit's state) and the residual (under the parent invoice's status): `deposit_i = ibl.itemizedAmount × (d_i.amount / I.amount)`, `residual = ibl.itemizedAmount × ((I.amount − Σd) / I.amount)`. Zero-deposit invoices behave identically to the pre-existing logic (regression-tested). All queries use one extra `LEFT JOIN invoice_deposits` — no N+1. Applies to budget overview, budget sources (paid/unclaimed/claimed/discretionary), and work-item + household-item budget summaries (`actualCost`/`actualCostPaid`/`actualCostClaimed`). No new schema, no new endpoints, no response-shape changes.

> **Note on mobile modal layout**: The UX spec mentioned a bottom-sheet modal on mobile. Per the dev-team-lead implementation spec, the shared center-modal is used across all viewports for this release (the bottom-sheet pattern is not yet part of the shared component library). A follow-up can revisit this once the UX designer specs the bottom-sheet as a reusable shared component — it is not a blocker for this feature.

> **Context**: #1403 (invoice deposits schema, CRUD API, cascade logic) already shipped. This PR completes the deposits feature end-to-end.

Fixes #1404
Fixes #1405

## Test plan
- [ ] Unit tests pass (95%+ coverage on new/modified files)
- [ ] Integration tests pass (`depositAggregateUtils`, `budgetServiceFactory`, `budgetSourceService`, `budgetOverviewService`)
[...]

Issue #1405, opened 2026-05-10
Invoice deposits: budget rollup integration (residual under parent status, deposits under their own state)


steilerDev:
**As a** homeowner reading any budget rollup (budget overview, budget categories, budget sources, subsidy claim totals, work-item / household-item budget summaries), **I want** an invoice's parent-residual portion to be bucketed under the parent's status while each deposit is bucketed under the deposit's own status, **so that** my paid/claimed/pending totals reflect what has actually been paid out and what has actually been reimbursed — even when a single invoice is staged across several deposits with different states.

**Parent Epic**: _none — see Story A. This is the budget-rollup story for the Invoice Deposits feature._
**Priority**: Should Have

## Background / Motivation

Story A delivers the `invoice_deposits` model and CRUD endpoints. Story B delivers the invoice-detail-page UI. Neither story changes how invoices contribute to budget rollups — both stories assume "Σ rollups continue to use the parent invoice's `amount` and `status` as today."

This story changes that. After this story merges:

- The portion of an invoice's amount that **is** allocated to deposits contributes to rollups under each deposit's own state (e.g., an invoice with `status: 'pending'` and one `paid` deposit and one `claimed` deposit produces three buckets: pending residual, paid deposit, claimed deposit).
- The **residual portion** (`amount − Σ deposits.amount`) — i.e., the "Final payment" — continues to contribute under the parent invoice's existing `status`.

The total contribution of a single invoice to a given budget line / source / category remains exactly equal to the invoice's `amount` — only the bucketing changes. This is a pure aggregation rewrite: no new schema, no new endpoints, no new UI.

## In Scope

1. All status-based budget rollups are updated to split each invoice's contribution into per-deposit buckets + a residual bucket bucketed under the parent's status. Identified rollup paths (the implementation spec must enumerate the complete list — these are the ones the PO has located in the codebase):
   - **Work-item budget summaries** (`shared/src/types/budget.ts` `actualPaid` / `actualClaimed` semantics; queries that aggregate `invoice_budget_lines` joined with `invoices.status`).
   - **Household-item budget summaries** (the parallel queries on the household-item side).
   - **Budget category aggregates** (used by `/budget/overview`).
   - **Budget source aggregates** (`shared/src/types/budgetSource.ts` `paidAmount` / `unclaimedAmount` / `claimedAmount`).
   - **Subsidy claim totals** (paid-and-claimed totals on subsidy pages).
[...]

PR #1412, opened 2026-05-12
fix(invoice): make status breakdown summary deposit-aware (#1411)

steilerDev:
## Summary

- `InvoiceStatusBreakdown.summary` (used by the InvoicesPage header cards and Dashboard InvoicePipelineCard) was not deposit-aware: a €1000 quotation invoice with a €200 pending deposit incorrectly showed €1000 under Quotation and €0 under Pending
- Adds `aggregateInvoiceStatusBreakdown()` to `depositAggregateUtils.ts` and rewrites `listAllInvoices()` summary to split amounts across parent + deposit statuses via LEFT JOIN on `invoice_deposits`; counts remain per-invoice (deposit rows never increment a count)
- Summary stays globally unfiltered (filter-independent) — consistent with existing UX; wiki API-Contract.md updated with the deposit-aware invariant and quotation example entry

Fixes #1411

## Test plan

- [ ] Unit tests for `aggregateInvoiceStatusBreakdown()` cover: no deposits, single deposit, multiple deposits with mixed statuses, deposit sum exceeding invoice amount (clamped to 0 residual)
- [ ] `invoiceService.test.ts` verifies deposit-aware summary in `listAllInvoices()` for global unfiltered totals
- [ ] Quality Gates CI passes (typecheck + unit tests + build)

### DIFF
--- a/server/src/services/invoiceService.test.ts
+++ b/server/src/services/invoiceService.test.ts
@@ -980,6 +980,153 @@ describe('Invoice Service', () => {
     });
   });
 
+  // ─── listAllInvoices() deposit-aware summary (#1411) ────────────────────────
+
+  describe('listAllInvoices() deposit-aware summary (#1411)', () => {
+    /**
+     * Insert a deposit directly into the DB (bypassing service validation).
+     * Scoped to this describe block; mirrors the helper in the #1403 block.
+     */
+    function insertRawDeposit(
+      invoiceId: string,
+      amount: number,
+      status: 'pending' | 'paid' | 'claimed' = 'pending',
+    ): string {
+      const id = `dep-summary-${Date.now()}-${Math.random().toString(36).substring(7)}`;
+      const ts = new Date(Date.now() + timestampOffset++).toISOString();
+      db.insert(schema.invoiceDeposits)
+        .values({
+          id,
+          invoiceId,
+          amount,
+          dueDate: '2026-03-01',
+          paidDate: status === 'paid' || status === 'claimed' ? '2026-02-15' : null,
+          claimedDate: status === 'claimed' ? '2026-02-20' : null,
+          description: null,
+          status,
+          createdBy: null,
+          createdAt: ts,
+          updatedAt: ts,
+        })
+        .run();
+      return id;
+    }
+
+    // ─── AC-1: quotation invoice with pending deposit ──────────────────────
+
+    it('AC-1: quotation invoice (1000) with pending deposit (200) — residual 800 under quotation, 200 under pending', () => {
+      const vendorId = createTestVendor('AC1 Vendor');
+      const invoiceId = insertRawInvoice(vendorId, {
+        // insertRawInvoice only types 'pending'|'paid'|'claimed' but the DB accepts 'quotation'
+        status: 'pending' as 'pending',
+        amount: 1000,
+      });
+      // Override status to 'quotation' via underlying better-sqlite3 (Drizzle type doesn't include it)
+      sqlite.prepare(`UPDATE invoices SET status = 'quotation' WHERE id = ?`).run(invoiceId);
+      insertRawDeposit(invoiceId, 200, 'pending');
+
+      const result = invoiceService.listAllInvoices(db, {});
+
+      expect(result.summary.quotation.totalAmount).toBe(800);
+      expect(result.summary.pending.totalAmount).toBe(200);
+      expect(result.summary.quotation.count).toBe(1);
+      expect(result.summary.pending.count).toBe(0);
+    });
+
+    // ─── AC-2: invoice with no deposits unchanged ──────────────────────────
+
+    it('AC-2 regression: invoice with no deposits has its full amount in the summary', () => {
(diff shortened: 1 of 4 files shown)


steilerDev:
**[product-architect]** Review: **APPROVED** with medium findings for refinement.

## Verdict

`APPROVED` — the fix is correct, complete, and resolves the gap that #1405 left behind. CI is green. Wiki is updated in the same PR (deposit-aware invariant documented on `summary.<status>.totalAmount`, `quotation` example added) — Wiki Update Discipline observed.

## Verified

- **Completes #1405's missed surface.** #1411 is precisely the rollup path that #1405's spec listed but the implementation skipped (`InvoiceStatusBreakdown.summary` consumed by `InvoicesPage` header cards + Dashboard `InvoicePipelineCard`). No other gap remains: the budget-line/source/category/work-item/household-item rollups were already migrated in #1407, and the per-invoice detail view (`getInvoiceById`) shows `deposits[]` directly rather than re-aggregating.
- **Math.** Invoice-level proportional split: `summary[I.status].totalAmount += max(0, I.amount − Σdeposits)`, `summary[deposit.status].totalAmount += deposit.amount`, count once per invoice. Defense-in-depth on negative residual via `Math.max(0,...)`. Sum invariant holds: `Σ summary[s].totalAmount === Σ I.amount` (Scenario 4 + integration test "sum invariant" both assert this).
- **Deduplication.** `deposits.some(d => d.depositId === row.deposit_id)` correctly handles the LEFT JOIN's row expansion (Scenario 8 explicitly tests this).
- **Global summary preserved (correct architectural call).** Pre-existing test at `invoiceService.test.ts:934` already pinned this: `summary reflects global invoice counts and totals across ALL invoices (ignores page filter)`. The wiki API-Contract.md line 3820 also documents it: `summary field is always the global unfiltered summary`. The PR preserves both. This deviates from #1411 AC-5 ('breakdown computed for the filtered set'), but the deviation is correct — the established UX is that header cards remain stable while filtering the list ('see what the other filter values would yield'), and the deposit-aware fix is orthogonal to filter scoping. AC-5 was inconsistent with the pre-existing system; the PR correctly chose the system's source of truth (existing test + wiki) over the issue's text.
- **Performance.** One LEFT JOIN over the full `invoices` table for the global summary. No N+1 (deduplication is in-memory by `Map`). At <5 users this is comfortably acceptable; even with thousands of invoices and deposits per project, this is one O(N+M) scan.
- **Test coverage.** 9 unit tests on `aggregateInvoiceStatusBreakdown` exercise empty input, no-deposits, single deposit, sum invariant, multi-deposit residual+accrual, clamp-to-zero, multi-invoice count, dedupe, and mixed scenarios. 6 integration tests cover AC-1..AC-8 against the live SQL. Coverage is 100% on the new helper.

## Medium findings (refinement-eligible, not blocking)

1. **Fourth instance of the row-grouping + deposit-dedup pattern.** `depositAggregateUtils.ts` now has three functions (`computeDepositAwareAggregates`, `computeStatusContribution`, `aggregateInvoiceStatusBreakdown`) that each re-implement the same `Map<invoiceId, ...> + depositsByInvoice + dedupe-by-depositId` preamble (lines 51-87, 170-205, 279-297) plus a `budgetSourceService.computeDiscretionaryInvoiceAmount` inline implementation. My review on #1407 already flagged this and recommended extracting a `splitByDeposits()` helper. This PR adds the fourth instance.

   **Recommendation:** Refactor in a follow-up — not in this PR. Reasoning: (a) the bug fix is small and focused, mixing it with a 3-call-site refactor would balloon the diff and risk breaking the working rollups; (b) the duplication is now well-isolated to one file with all four call sites visible together, which is the right shape for a clean extraction; (c) no tracking issue exists yet — please open one (`refactor(invoice): extract splitByDeposits helper from depositAggregateUtils`) and tag it for refinement during `/epic-close` or the next cleanup pass. The right abstraction is a single function that takes the raw rows + a per-invoice callback `(invoiceAmount, parentStatus, deposits[]) => void`, leaving the per-call-site accumulation logic to the caller.

2. **Two minor nits in the new helper** (not blocking, can fold into the refactor above):
   - `ensure(deposit.depositStatus)` is called even for deposits whose status equals the parent invoice's status — no functional issue (idempotent), just a tiny waste.
   - The interface name `InvoiceDepositRow` is fine but consider `InvoiceWithDepositRow` to disambiguate from `invoice_deposits` table rows.

3. **Test comment cleanliness.** `invoiceService.test.ts` AC-1 uses a raw SQL `UPDATE` to set `status='quotation'` because `insertRawInvoice` doesn't accept `'quotation'` in its typed union. Consider widening that helper's status param to the full `InvoiceStatus` union in a follow-up — it would clean up this and any future quotation test cases. Not a blocker.

## Not findings

- The global-stays-global decision is correct and consistent with the existing system. The wiki, the pre-existing test, and this PR all align.
- Performance concerns at scale are theoretical at this project's <5-user target; not worth optimizing.

Architecturally sound. Approving — please open the refactor tracking issue before merge so the duplication doesn't get forgotten.

Issue #1435, opened 2026-05-17
BUG: Diary UX rough edges after #1426 — auto-draft on type select, photo grid refresh, status filter


steilerDev:
**[product-owner]**

Three UX rough edges introduced or surfaced by the auto-draft + immediate-upload flow shipped in #1426. All three are client-only changes — no API or schema work needed.

## Problem

After #1426 the diary draft and photo flow works end-to-end, but day-to-day use surfaced three rough edges:

1. The create flow has an unnecessary intermediate \"form-but-no-draft\" state where a user has picked a type but no draft exists yet.
2. Photo uploads succeed but the new photo only appears after a page refresh, so the user has no immediate confirmation that the upload landed.
3. The new status filter is rendered as a standalone three-chip row (All / Drafts only / Saved only) sitting between the filter bar and the timeline. That is far too prominent for what is effectively a single boolean.

## Expected behavior

- **Create flow** — Clicking a type card on `/diary/new` immediately creates the draft and navigates to `/diary/:id/edit`. No intermediate form state on the create page.
- **Photo upload** — When `PhotoUpload`'s `onUpload` fires on the edit page, the new photo is visible in the entry's `PhotoGrid` within ~500ms without a page reload.
- **Status filter** — The standalone status chip row is gone. A single \"Hide drafts\" toggle lives inside the existing `DiaryFilterBar` filter header. Default off (drafts visible). On → applies `?status=saved`; off → no status filter in the URL.

## Actual behavior

- **Create flow** — `client/src/pages/DiaryEntryCreatePage/DiaryEntryCreatePage.tsx` uses a two-step `'type-selector' | 'form'` state machine. `handleTypeSelect` (line 134) just flips `step` to `'form'`. The draft is only created later on the first meaningful metadata change (effect at line 111) or on first field blur via `onFieldBlur={createDraft}`. The user sees a form rendered locally with no `entryId` yet — exactly the intermediate state #1426 was meant to remove.
- **Photo upload** — `DiaryEntryEditPage` (line 698) passes `onUpload={() => { /* Photo is automatically added to state by usePhotos */ }}`. That comment is wrong: `PhotoUpload` calls `uploadPhoto` from `photoApi` directly (line 104 of `PhotoUpload.tsx`), not `usePhotos.uploadPhoto`, so `usePhotos` never sees the new photo and never updates its `photos` state. The grid only refreshes when the page remounts.
[...]

Issue #1426, opened 2026-05-15
BUG: Diary photos lost on upload failure — replace local-stage flow with auto-draft + immediate upload


steilerDev:
## Problem

When a user attaches photos on the diary entry **create** page, the photos are buffered locally in component state (`pendingFiles`) and only uploaded to the server **after** the entry has been POSTed via `POST /api/diary-entries`. If any of those subsequent uploads fail (network blip, file too large, server error, browser tab closed mid-upload, etc.), the user's photos are **silently lost**: the entry has been created but without the photos, and the form has been navigated away from. The user has no way to re-attach the lost photos because the original `File` objects are gone from memory.

A real user reported losing many photos this way. The frustration is amplified for diary entries because diary photos are typically taken on-site once and cannot be retaken.

The same problem affects other "fail late, lose data" surfaces on the create form: the user can fill out a long body, add a signature, and queue 10 photos, and a single network failure on submit destroys everything.

## Current behaviour (today)

Investigated in `client/src/pages/DiaryEntryCreatePage/DiaryEntryCreatePage.tsx`:

- `pendingFiles` is a local `useState<File[]>([])` — purely in-memory, never persisted server-side.
- The form does **not** create a server resource until the user clicks "Save". All field state (title, body, metadata, signatures) lives only in React state.
- On submit (`handleSubmit`), the flow is:
  1. `await createDiaryEntry(...)` — POSTs the entry, returns `entry.id`.
  2. `await Promise.all(pendingFiles.map((file) => uploadPhoto('diary_entry', entry.id, file)))` — uploads all photos in parallel against the just-created entry.
  3. On success: toast + `navigate(/diary/${entry.id})`.
  4. On photo upload failure: a generic toast "photo upload failed" is shown, but the code still falls through to the success toast and navigates away. The `File` objects in `pendingFiles` are discarded with the unmounted component.
- If the user closes the browser tab or loses connection between steps 1 and 2, the entry exists with zero photos and the photos are lost.
- If the user closes the browser tab before submitting at all, **everything** is lost (body, metadata, signatures, queued photos).

By contrast, the **edit** page (`DiaryEntryEditPage.tsx`) already does the right thing: photos are uploaded immediately to the existing entry id via `<PhotoUpload entityType="diary_entry" entityId={entry.id} ... />`. The bug is specific to the create flow because no server-side entry exists yet to attach photos to.

## Expected behaviour (new flow)

[...]

PR #1436, opened 2026-05-17
fix(diary): immediate draft on type-card click; photo refresh; hide-drafts toggle (#1435)

steilerDev:
## Summary

Three UX polish items for the diary feature after #1426 landed:

- **AC1** — Clicking a type card on `/diary/new` now immediately creates the draft and navigates to `/diary/:id/edit`. The intermediate form step is gone — all editing happens on the edit page where it already worked.
- **AC2** — Newly uploaded photos appear in the photo grid right away. `PhotoUpload.onUpload` now calls `usePhotos.refresh()` instead of being a no-op.
- **AC3** — The standalone "All / Drafts only / Saved only" chip row on `/diary` is replaced by a single "Hide drafts" checkbox in the existing filter bar. Toggle off (default): drafts visible. Toggle on: `?status=saved`.

Frontend-only — no API, schema, or backend changes.

Fixes #1435

## Test plan

- [x] Unit: DiaryFilterBar 38/38 local; new scenarios added to DiaryPage, DiaryEntryCreatePage, DiaryEntryEditPage (local fails are the documented sandbox mock issue — pass in CI)
- [x] E2E: diary-drafts.spec.ts scenarios 1-4, 6, 12-14 updated; diary-forms.spec.ts and diary-uat-fixes.spec.ts migrated for the type-card flow
- [x] i18n: en/de parity preserved; new key `filterBar.hideDrafts` ("Entwürfe ausblenden")


steilerDev:
**[product-architect]** Architecture review

Verdict: **APPROVED WITH NOTES** (one real bug worth fixing before merge, but not architectural)

### Scope check
- Frontend-only confirmed: zero changes to `server/`, `shared/`, or `wiki/`. No schema/API drift. Good.

### AC1 — DiaryEntryCreatePage simplification
- Clean removal: 339 lines → 56 lines. `useEffect` for vendors, all metadata state, `buildMetadata`, `validateForm`, `handleFileSelect`, the `Step` machine, and the `useAuth` dependency are all gone. No abandoned hooks left behind.
- `draftCreatingRef` + `isCreating` state combine well for the double-click guard (ref blocks reentry synchronously; state disables cards visually). The `void handleTypeSelect(...)` wrappers correctly preserve the promise lifecycle.
- Note: on error path the cards are re-enabled (`setIsCreating(false)`) but `error` state is no longer rendered to the UI (`errorBanner` CSS class was deleted). Failure is communicated solely via toast, which is fine — but the `error` local state would have been dead code if not also removed. Confirmed it was removed.

### AC2 — Photo refresh via `onUpload`
- `onUpload={() => photosResult.refresh()}` is the right call here. The optimistic path would require threading the just-uploaded `Photo` into `usePhotos` state, which the hook doesn't currently expose as a mutator. Re-fetch is simpler, the network cost is one cheap list call, and #1 user scale makes optimistic UI not worth the added surface area. Approve as-is.

### AC3 — `DiaryFilterBar` hideDrafts checkbox
- Backward-compatible: both props optional, conditional render (`{onHideDraftsChange && ...}`), checkbox only appears when integration opts in. Existing call sites (e.g., dashboard mini-view) remain unaffected.
- URL contract preserved: `?status=saved` is still the persisted state; `hideDrafts` is a derived view on top (`hideDrafts = statusFilter === 'saved'`). This means deep links from before #1435 continue to work, and the only loss is the ability to filter to drafts-only via the UI (still possible via `?status=draft` URL). E2E POM confirms via `filterDraftsOnly()` helper. Acceptable trade for the UX simplification.
- The `#hide-drafts-toggle:focus-visible` selector targets a literal ID via CSS Modules — this works but is unusual (ID selectors bypass the module's class hashing). Not a blocker; consider moving the focus style to a `.hideDraftsCheckbox` class for consistency with the rest of the file.

### One real bug — dangling i18n keys

The PR removes these keys from both `en/diary.json` and `de/diary.json`:
- `createPage.siteVisitInspectorNameRequired`
- `createPage.issueValidationRequired.severity`
- `createPage.issueValidationRequired.resolutionStatus`

But `client/src/pages/DiaryEntryEditPage/DiaryEntryEditPage.tsx` still references them on lines 304, 313, 316 in its `validateForm()` for site_visit and issue entries. When a user clicks Promote on a site_visit draft without an inspector name (or an issue draft without severity/resolution), the validation error will fall back to the raw key string instead of the localized message.

Fix options (pick one — frontend-developer):
1. Move the three keys into the `edit.*` namespace (e.g., `edit.siteVisitInspectorNameRequired`) since the validation now lives exclusively on the edit page, and update the three `t()` calls.
2. Keep them as `createPage.*` in the JSON files (the cleanup overreached).

Option 1 is the cleaner outcome — the create page no longer has type-specific validation at all, so `createPage.*` validation keys are misplaced.

### Architectural debt / TODOs
- None flagged. The `// NOTE: bumped to 15000ms; underlying lazy-import flake tracked in #1438.` annotation in `App.test.tsx` is properly tracked.
- E2E POM comments are well-maintained and explain the post-#1435 flow clearly for future agents.

### Summary
Architecture and contracts are sound. Approving on the architecture dimensions. The dangling i18n keys are a frontend bug, not an architecture concern — frontend-developer should fix before merge, but it doesn't block the architectural design.

Co-Authored-By: Claude product-architect (Opus 4.6) <noreply@anthropic.com>

steilerDev (commented):
**[ux-designer]** Design review

**Verdict: APPROVED**

Reviewed three areas: `DiaryFilterBar.module.css` (new checkbox rules), `DiaryEntryCreatePage.module.css` (form-step removal cleanup), `DiaryPage.module.css` (chip-row deletion).

---

### Token adherence

All values in the new `.hideDraftsRow` / `.hideDraftsLabel` / `#hide-drafts-toggle:focus-visible` rules use design tokens:

- `gap: var(--spacing-2)` ✓
- `font-size: var(--font-size-sm)`, `font-weight: var(--font-weight-medium)`, `color: var(--color-text-primary)` ✓
- Focus ring: `outline: none; box-shadow: var(--shadow-focus)` — correct pattern ✓

`DiaryEntryCreatePage.module.css` — the previously-flagged `transition: all 0.2s` is now `var(--transition-medium)` on the PR branch. Confirmed on `origin/fix/1435-diary-ux-polish`. Dead form-step rules cleanly removed.

`DiaryPage.module.css` — all three chip classes deleted cleanly, no dead CSS left behind.

---

### Checkbox a11y

`<label htmlFor="hide-drafts-toggle">` paired with `id="hide-drafts-toggle"` on the `<input>`. Label precedes the input in DOM order — conventional and correct. `data-testid="hide-drafts-checkbox"` present. Accessible label verified by test Scenario 17 (`getByLabelText(/hide drafts/i)`). ✓

---

### Touch target

The label and input share a flex row (`align-items: center`). The label has `cursor: pointer` so the full label text is a tap target. Native checkboxes are small (~16–20 px), but the label text extends the tap area substantially. Adequate for the filter-bar context (this is a secondary filter control, not a primary action). ✓

---

### Layout / grid placement

`.hideDraftsRow` uses `grid-column: 1 / -1` — spans the full three-column grid, placing it as a full-width row below the type chips and above the clear button. Consistent with how `.filterGroup:nth-child(5)` and `:nth-child(6)` span full width. No grid breakage. ✓

---

### Dark mode

All color references (`--color-text-primary`, `--shadow-focus`) are semantic tokens defined in `[data-theme="dark"]` in `tokens.css`. No hardcoded colors introduced anywhere in the diff. ✓

---

### One informational note (non-blocking)

The global ID selector `#hide-drafts-toggle:focus-visible` in a CSS Module file is atypical — CSS Modules scope class selectors but ID selectors are not scoped. This works correctly (there will only ever be one instance of `DiaryFilterBar` on screen at a time), but it is worth knowing the ID escapes module scoping. A class-based approach (e.g., adding a `.hideDraftsToggle` class to the `<input>` and styling that) would be more idiomatic for CSS Modules. Flag for awareness, not a blocker.

Issue #1475, opened 2026-05-17
Photo annotator: geometric tools (Arrow, Line, Ellipse)


steilerDev:
**As a** household user, **I want** to draw arrows, straight lines, and ellipses on photos with optional shift-constrain shortcuts, **so that** I can point at, outline, and circle features in construction photos with the same precision and ergonomics as the Rectangle tool.

**Parent Epic**: #1472
**Priority**: Should Have

### Scope

Adds three geometric drawing tools to the annotator built in Story 1. Each follows the same draft-shape lifecycle (pointerdown starts draft, pointermove updates, pointerup commits) and integrates with Select/move/resize/delete + undo/redo from Story 1.

- \`client/src/components/photos/PhotoAnnotator/tools/ArrowTool.ts\` — SVG \`<marker>\` arrowhead plumbing in \`render.ts\` for both SVG and canvas-2d rendering paths
- \`client/src/components/photos/PhotoAnnotator/tools/LineTool.ts\` — Shift held constrains the line to horizontal, vertical, or 45° (snap to nearest)
- \`client/src/components/photos/PhotoAnnotator/tools/EllipseTool.ts\` — Shift held constrains to a circle (\`rx === ry\`)
- Shape definitions extended in shape types union: \`{ type:'arrow', x1,y1,x2,y2, stroke, strokeWidth }\`, \`{ type:'line', x1,y1,x2,y2, stroke, strokeWidth }\`, \`{ type:'ellipse', cx,cy,rx,ry, stroke, strokeWidth, fill? }\`
- Tool palette icons (added to \`ToolPalette.tsx\`) with translated labels (\`photoAnnotator.tool.arrow\`, \`.line\`, \`.ellipse\`)
- Renderer updates in \`render.ts\` to draw arrow + arrowhead, line, and ellipse on both SVG and canvas-2d so the bake matches the on-screen preview pixel-for-pixel
- Co-located unit tests for each new tool and the renderer extensions

### Acceptance Criteria

- [ ] **Given** the annotator is open, **when** the user selects the Arrow tool and drags from point A to point B, **then** an arrow shape appears tracking the drag with an arrowhead at the destination, and commits on pointerup.
- [ ] **Given** the annotator is open, **when** the user selects the Line tool and drags, **then** a straight line appears tracking the drag and commits on pointerup.
[...]

PR #1486, opened 2026-05-18
feat(photo-annotator): add geometric tools (arrow, line, ellipse)

steilerDev:
## Summary

Implements three new geometric drawing tools for the photo annotator with full support for drawing, selection, moving, and resizing:

- **Arrow tool** — draws arrows with automatic arrowhead markers, 2px minimum length
- **Line tool** — draws lines with Shift-constrain to 45° angles, 2px minimum length
- **Ellipse tool** — draws ellipses with Shift-constrain to circles, 1px minimum radii

All shapes support full editing: move by dragging the body, resize via endpoint handles (arrow/line) or cardinal handles (ellipse). Keyboard nudge (arrow keys) works for all shape types.

## Changes

**Files Created:**
- `client/src/components/photos/PhotoAnnotator/tools/ArrowTool.ts` — arrow drawing tool
- `client/src/components/photos/PhotoAnnotator/tools/LineTool.ts` — line drawing with 45° snapping
- `client/src/components/photos/PhotoAnnotator/tools/EllipseTool.ts` — ellipse drawing with circle constraint

**Files Modified:**
- `client/src/components/photos/PhotoAnnotator/useUndoStack.ts` — extended `AnnotationShape` union with `ArrowShape`, `LineShape`, `EllipseShape`
- `client/src/components/photos/PhotoAnnotator/useAnnotator.ts` — added new shape types to `ToolName` union and re-exports
- `client/src/components/photos/PhotoAnnotator/geometry.ts` — added 8 new geometry helpers for hit-testing and transforms
- `client/src/components/photos/PhotoAnnotator/render.ts` — added SVG and canvas-2d rendering for new shapes
- `client/src/components/photos/PhotoAnnotator/tools/SelectTool.ts` — extended hit-testing and move/resize dispatch for all shape types
- `client/src/components/photos/PhotoAnnotator/PhotoAnnotator.tsx` — integrated new tools, added SVG arrowhead marker, extended selection overlay with endpoint/cardinal handles, updated keyboard nudge
- `client/src/components/photos/PhotoAnnotator/ToolPalette.tsx` — added three tool buttons with ArrowIcon, LineIcon, EllipseIcon SVG components
- `client/src/components/photos/PhotoAnnotator/tools/RectangleTool.ts` — cast draft shape for type safety
- `client/src/components/photos/PhotoAnnotator/tools/HighlightTool.ts` — cast draft shape for type safety
- `client/src/i18n/en/photoAnnotator.json` — added 6 English translation keys

## Architecture Notes

- **Marker design:** SVG arrowhead uses `fill="context-stroke"` so it inherits the line's stroke color without dark-mode overrides
- **Handle types:** Arrows/lines use `start`/`end` handles; ellipses use cardinal `north`/`south`/`east`/`west` handles; rectangles/highlights use 8-point box handles
- **Shift constraints:** Line snaps to nearest 45° angle; ellipse constrains to circle (equal radii)
- **Stroke width:** Uses same `{ thin: 2, medium: 4, thick: 8 }` map as rectangles
- **No hardcoded colors:** All shapes use semantic colors from palette; arrowhead uses `context-stroke`

## Test Notes

Production code compiles with strict TypeScript. Test files have minor union-type issues that the QA agent will fix.

No translation to German — translator owns non-English locales.

Closes #1475


steilerDev (commented):
**[ux-designer]** Design compliance review for PR #1486 — Photo Annotator Geometric Tools (Story #1475)

---

## Summary

Reviewed: `ToolPalette.tsx`, `PhotoAnnotator.tsx`, `render.ts`, `ToolPalette.module.css`, `en/photoAnnotator.json`, `de/photoAnnotator.json`.

Overall the implementation is solid and closely follows the patterns established in Story #1483. All findings are **low/informational** — nothing blocks merge.

---

## What Passes ✓

**Token adherence (CSS Modules):** `ToolPalette.module.css` uses design tokens exclusively (`var(--spacing-*)`, `var(--color-*)`, `var(--radius-*)`, `var(--transition-*)`, `var(--shadow-focus)`). No hardcoded values in the stylesheet. No new CSS module files were added — correct, the existing `.toolButton` class covers the new buttons without changes.

**SVG inline color values (intentional pattern):** The `stroke:` values on committed shapes (e.g. `'#dc2626'`, `'#3b82f6'`, `'#16a34a'`) are user-selected annotation colors from `ANNOTATION_COLORS`. These are intentionally theme-invariant hardcoded hex — marks must look the same regardless of theme. This is documented in agent memory as the established pattern. Not a violation.

**Arrowhead dark mode:** `fill="context-stroke"` on the `<marker>` polygon is correct SVG2 — the arrowhead inherits the line's stroke color, which is itself a user-selected annotation color. No `[data-theme="dark"]` override needed. Well done.

**Selection overlay tokens:** All selection overlays (`stroke="var(--color-primary)"`, `fill="var(--color-bg-primary)"`, `stroke="var(--color-primary)"`) use semantic tokens. These will flip correctly in dark mode.

**Draft shape visual:** Arrow and line: `stroke-dasharray: '6 4'`, `opacity: 0.8`, `pointer-events: none` on draft. Ellipse: same treatment. Matches the spec pattern exactly.

**Tool button accessibility:** All three new buttons have `aria-pressed`, `data-testid`, `aria-label={t('toolArrow|Line|Ellipse')}`, and use the `.toolButton` / `.toolButtonActive` CSS classes. `min-width/height: 44px` is inherited from the class. Focus ring via `box-shadow: var(--shadow-focus)` on `:focus-visible`. Correct.

**Icons:** `ArrowIcon`, `LineIcon`, `EllipseIcon` — all 24×24, `stroke="currentColor"`, `strokeWidth="2"`, `fill="none"`, consistent with `RectangleIcon`, `UndoIcon`, `RedoIcon`. The `ArrowIcon` correctly uses two path segments (line + head) to convey directionality in the toolbar.

**Responsive (mobile):** No new CSS was needed — the existing `@media (max-width: 639px)` block gives `.toolGroup` `width: 100%`. The three new buttons sit inside `.toolGroup` and inherit this. Correct per the pattern.

**Handle counts per shape type:** Ellipse → 4 cardinal (N/S/E/W). Arrow/Line → 2 endpoint handles. Rectangle/Highlight → 8 corner+edge handles. All match the specified patterns.

**i18n:** `en/photoAnnotator.json` has `toolArrow`, `toolLine`, `toolEllipse`, `shapeAddedArrow`, `shapeAddedLine`, `shapeAddedEllipse`. `de/photoAnnotator.json` has all corresponding keys with correct German compounds (`Pfeil-Werkzeug`, `Linien-Werkzeug`, `Ellipsen-Werkzeug`). Parity confirmed.

**`prefers-reduced-motion`:** All transitions in `ToolPalette.module.css` are wrapped in `@media (prefers-reduced-motion: no-preference)`. No motion added by this PR outside that guard.

---

## Informational Notes (non-blocking)

**1. Arrowhead on draft omitted correctly, but `'none'` string for `marker-end`**

In `render.ts` line ~60:
```ts
'marker-end': isDraft ? 'none' : 'url(#arrowhead)',
```
The SVG attribute value `'none'` is valid for `marker-end` — this will suppress the arrowhead on draft correctly. No change needed, just noting it reads as a string rather than the keyword to ensure reviewers don't confuse it for the CSS `display: none` pattern.

**2. Ellipse handle cursor on `'west'` and `'east'`**

The cursor is correctly `'ew-resize'` for east/west and `'ns-resize'` for north/south. If the ellipse is very wide (large `rx`), east/west handles are at the ellipse equator — `ew-resize` is semantically correct there. No change needed.

**3. SVG `<marker>` `id="arrowhead"` is hardcoded**

The marker id is a plain string `"arrowhead"`. This works fine in a single-instance annotator. If `PhotoAnnotator` were ever rendered more than once on the same page, the second SVG's `url(#arrowhead)` would resolve to the first instance's marker. This is a low-risk future concern for this self-contained component — worth a code comment noting the assumption of single-instance.

**4. `ToolPalette.module.css` hover state uses `[aria-pressed='true']` guard but new buttons use `aria-pressed={boolean}`**

The CSS selector `.toolButton:hover:not([aria-pressed='true'])` uses the string `'true'`. React serializes `aria-pressed={true}` as the attribute `aria-pressed="true"` in the DOM — this matches correctly. Just confirming the boolean-prop-to-string-attribute serialization works as expected here.

---

## Verdict

**APPROVED.** No design system violations. Token adherence is correct. Icon style is consistent. Accessibility attributes are complete. Dark mode handling is sound. Responsive behavior is inherited correctly from the existing toolbar layout.

Informational notes above are for awareness only and do not require changes before merge.

steilerDev:
**[product-owner]**

## Verdict: CHANGES_REQUIRED

Reviewed PR #1486 against Story #1475 acceptance criteria. 8 of 9 criteria pass. **AC #4 fails for 3 of 4 drag directions due to an EllipseTool bounding-box math bug.**

(Posted as a comment because GitHub disallows `--request-changes` on a self-authored PR. Please treat this as a blocking review.)

## AC Check

| # | Criterion | Status |
|---|---|---|
| 1 | Arrow tool: drag draws arrow with arrowhead at destination, commits on pointerup | PASS — `ArrowTool.ts:13-72`, marker plumbed in `PhotoAnnotator.tsx:374-386` and `render.ts:51-67` |
| 2 | Line tool: drag draws straight line, commits on pointerup | PASS — `LineTool.ts:35-106` |
| 3 | Line tool + Shift: snaps to nearest 0°/45°/90°/etc. | PASS — `LineTool.ts:15-33,75-79` (nearest of 8 angles via `atan2` + π/4 rounding) |
| 4 | Ellipse tool: drags out a bounding box, ellipse **fits that bounding box** | **FAIL** — see below |
| 5 | Ellipse tool + Shift: constrains to circle (`rx === ry`) | PASS — `EllipseTool.ts:53-58` |
| 6 | Select tool: click selects, drag moves, handles resize, Delete/Backspace removes (for all three new shapes) | PASS — endpoint handles for arrow/line (`PhotoAnnotator.tsx:471-501`), cardinal handles for ellipse (lines 503-536), wired in `SelectTool.ts:46-79` |
| 7 | Undo/redo removes/restores arrow, line, ellipse | PASS — all three are part of the `AnnotationShape` union flowing through the existing undo stack (`useUndoStack.ts`) |
| 8 | Save: baked PNG matches on-screen SVG preview (arrowhead, straight line, ellipse curvature) | PASS — `render.ts:113-157` mirrors SVG output for canvas-2d |
| 9 | Jest unit tests pass; 95%+ coverage on each new tool + extended renderer | PASS (pending CI green) — test files exist for all three tools, geometry helpers, and render extensions |

## Blocking Issue — AC #4

**`client/src/components/photos/PhotoAnnotator/tools/EllipseTool.ts:47-66`** computes the ellipse center incorrectly when the user drags in any direction other than down-and-right:

```ts
const dx = Math.abs(imageX - drawState.startX);
const dy = Math.abs(imageY - drawState.startY);
// ...
cx: drawState.startX + dx / 2,
cy: drawState.startY + dy / 2,
```

Because `dx`/`dy` are absolute values but `cx`/`cy` are computed by adding (never subtracting) from `startX`/`startY`, the ellipse center always lands down-right of the start point.

**Reproduction (numeric):**
- Start at (200, 200), drag pointer to (100, 120) — bounding box should be (100,120)–(200,200), center (150, 160).
- Current code: `dx=100, dy=80, cx = 200 + 50 = 250, cy = 200 + 40 = 240, rx=100, ry=80`.
- Rendered ellipse spans (150,160)–(350,320) — fully outside the user's drag rectangle, far to the lower-right.

This breaks AC #4 for the up-left, up-right, and down-left drag quadrants (only down-right works). For comparison, `RectangleTool.ts:46` correctly uses `normalizeRect(...)` for the same problem.

**Secondary issue (also breaks AC #4 even in the down-right quadrant):** the current code uses full `dx`/`dy` as radii. The AC says \"an ellipse fitting that bounding box\" — the radii must be **half** the bounding-box width/height so the ellipse touches all four sides. Today, dragging a 100×60 box renders a 200×120 ellipse.

**The existing unit test at `EllipseTool.test.ts:196-210`** (`'produces positive rx/ry even when dragging to top-left of start point'`) only asserts `rx`/`ry`, not `cx`/`cy`, so it silently passes the bug. The test needs to add `cx`/`cy` assertions covering all four drag directions, and `rx`/`ry` assertions need to reflect the half-bounding-box semantics.

## Suggested Fix

Replace the center + radii computation with the bounding-box midpoint:

```ts
const minX = Math.min(drawState.startX, imageX);
const minY = Math.min(drawState.startY, imageY);
const dx = Math.abs(imageX - drawState.startX);
const dy = Math.abs(imageY - drawState.startY);

let rx = dx / 2;
let ry = dy / 2;
if (event.shiftKey) {
  const r = Math.max(rx, ry);
  rx = r;
  ry = r;
}

const updatedDraft: EllipseShape = {
  ...shape,
  cx: minX + dx / 2,
  cy: minY + dy / 2,
  rx,
  ry,
};
```

The `rx>=1 && ry>=1` commit threshold can stay; with the divide-by-2 it now means a 2×2 bounding box minimum, consistent with the 2px minimum on Rectangle/Arrow/Line.

## Test Coverage Gaps to Address

Add to `EllipseTool.test.ts`:
- Drag up-left, up-right, down-left, down-right — assert `cx`/`cy` match the bounding-box midpoint in all four cases.
- Bounding-box fit — for a drag from (100,100) to (300,200), assert `rx === 100`, `ry === 50`, `cx === 200`, `cy === 150`.
- Shift constraint when dragging up-left — assert center stays at midpoint and `rx === ry === max(dx, dy)/2`.

Issue #1600, opened 2026-05-26
[auto-itemize] Assignment dialog: suggestion overflow, missing create-new affordance, and no auto-create from accepted itemization


steilerDev:
**[product-owner]**

The "assign to work item / household item" dialog launched from the Auto-Itemize page (`AutoItemizePage`, lines ~1223-1382) has three related defects when assigning an extracted line item to a parent (work item or household item) and choosing or creating its linked budget line. The first is a UX/clipping issue, the second blocks the user entirely when the parent already has budget lines, and the third forces redundant manual data entry when the information is already extracted by the LLM.

## Affected code

- `client/src/pages/AutoItemizePage/AutoItemizePage.tsx` — modal markup at lines 1223-1382; row-data shape and `onLineCreated` callback at lines 111-140.
- `client/src/components/ParentPicker/ParentPicker.tsx` + `.module.css` — step-1 tabs wrapping `WorkItemPicker` / `HouseholdItemPicker`.
- `client/src/components/SearchPicker/` — renders the suggestion dropdown clipped in flaw 1.
- `client/src/hooks/useBudgetLinePicker.ts` — `showCreateBudgetLineForm` (lines 188-231) initializes the create form with empty strings, ignoring the auto-itemized row context.
- `client/src/components/Modal/Modal.tsx` and `AutoItemizePage.module.css` (`.pickerContent`, `.pickerStep`) — overflow constraints causing the clip.
- `shared/src/types/budgetExtraction.ts::ExtractedLine` — source of the data that should pre-fill the create form (description, totalAmount, quantity, unit, unitPrice, includesVat, vatRate, confidence, vendorName, budgetCategoryId, budgetSourceId).

### Flaw 1 — Suggestion dropdown clipped by modal bounds

The `ParentPicker` (step 1 of the modal) renders `WorkItemPicker` / `HouseholdItemPicker`, which use a `SearchPicker` whose suggestion list is absolutely positioned beneath the input. The `Modal` body and the surrounding `.pickerContent` / `.pickerStep` containers clip the dropdown so the user cannot see the full list of matches.

- **Expected**: the suggestion list flows outside the modal bounds (overlay/popover-style) so all results are visible regardless of modal height.
- **Actual**: the dropdown is truncated at the modal's content edge; matches below the cut-off are unreachable except via continued typing.
[...]

PR #1601, opened 2026-05-26
fix(auto-itemize): assignment dialog overflow, create-new affordance, auto-prefill from extraction

steilerDev:
## Summary

Fixes three flaws in the auto-itemize assignment dialog:

- **Suggestion dropdown overflow**: `SearchPicker` dropdown now portals to `document.body`, repositions on scroll/resize, flips above input when no space, and closes on Escape from either the input or the portal.
- **"Create Budget Line" button always visible**: previously only rendered in the empty-state branch of step 2; now appears whether the parent has zero or many existing budget lines.
- **Auto-prefill from accepted itemization**: when the user clicks "Create Budget Line" from a row, the new form is pre-populated with the row's already-extracted properties (description, total amount, confidence, category, source, vendor, quantity/unit/unit price, includes-VAT). A new "Auto-created" badge surfaces on the row's assigned-line summary so the provenance is clear. For household-item parents, `budgetCategoryId` is omitted.

Fixes #1600

## Test plan

- [ ] Unit tests pass (95%+ coverage on modified files in CI)
- [ ] Integration tests pass
- [ ] E2E scenarios 33/34/35 pass (`invoice-auto-itemize-page.spec.ts`)
- [ ] Quality Gates green


steilerDev (commented):
**[ux-designer]**

Reviewed `Badge.module.css`, `SearchPicker.module.css`, `AutoItemizePage.module.css`, `AutoItemizePage.tsx`, and the PR diff. Results below.

---

### 1. Badge `.info` variant — token chain and semantic fit

Token choices are correct: `--color-bg-tertiary` / `--color-text-secondary` / `--color-border` all resolve cleanly in both light and dark mode via the existing Layer 3 overrides (dark: `--color-slate-600` bg / `--color-slate-150` text / `--color-slate-500` border). Contrast in dark mode is approximately 5.3:1 between `--color-slate-150` (#cbd5e1) on `--color-slate-600` (#1e293b) — passes WCAG AA for small text.

The variant name `.info` is appropriately generic and is a good fit for the shared Badge library. It does not overlap with any existing variant. The comment `/* Info badge (e.g., "Auto-created") */` is helpful. The `badgeStyles.info` class is correctly wired in `createdFromExtractionVariants` with both `label` and `className` set — no repetition of the recurring UNASSIGNED_BADGE_VARIANTS bug.

One note on **semantic fit**: the "Auto-created" badge signals a successful automated action. A `--color-success-*` treatment (green tint) would communicate that semantics more precisely. However, the neutral `.info` style is also defensible — it avoids false-positive "success" signals in cases where auto-creation was a fallback. The current implementation is acceptable as-is. If a success-tinted "system action" variant is needed elsewhere in the future, this can be revisited as a style guide refinement.

---

### 2. Token adherence — all new CSS rules

- `Badge.module.css .info` — three properties, all tokens. Clean.
- `SearchPicker.module.css .portalDropdown` — all tokens. Clean.
- `AutoItemizePage.module.css .assignedBadgeWrapper` — `gap: var(--spacing-2)`, all tokens. Clean.

No hardcoded values in the diff.

---

### 3. Portal dropdown styling parity

`.portalDropdown` and `.dropdown` share identical visual tokens: `--color-bg-primary` background, `--color-border-strong` border, `--radius-md`, `--shadow-md`, `max-height: 300px` / mobile `250px`, `overflow-y: auto`. The only intentional delta is `position: fixed` (set via inline style) and the elevated z-index — both required by the portalling approach. Visual parity is confirmed.

The `z-index: calc(var(--z-modal) + 100)` arithmetic is necessary but is a medium-level concern: it is a raw calc on a token rather than a named token. This is the correct approach given the current token scale (`--z-modal: 1000`; no higher named level exists). No action required for this PR; flag for token scale extension in a future refinement if more overlay-above-modal surfaces emerge.

---

### 4. Dark mode

All new color usages are semantic tokens that switch in `[data-theme="dark"]`. No dark mode regressions.

---

### 5. Responsive — "Create Budget Line" button touch target

The button uses `.addButton` (padding `var(--spacing-2) var(--spacing-3)` with `font-size: var(--font-size-sm)`). Computed height: 2×8px padding + ~21px line height ≈ 37px. This falls below the 44px minimum touch target. This button is rendered inside a modal, which is typically accessed on tablet/desktop — mobile use is unlikely but the WCAG 2.5.5 target size guideline applies regardless. Non-blocking; add `min-height: 44px` to `.addButton` before merge or in refinement.

The `max-height: 250px` on mobile for `.portalDropdown` is reasonable.

---

### 6. Accessibility — portal + focus trap

The Escape handler on the input (`onKeyDown` in the `<input>` element) and the matching handler on the portal's `<div role="listbox">` cover both entry points cleanly. The focus trap inside the Modal is not broken: the portal is rendered into `document.body` but the listbox is reached via keyboard from the focused input inside the modal — Tab/arrow navigation stays within the modal's natural flow. The portal node sits outside the trap boundary but is only reachable from within it, so there is no escape path. This is architecturally sound.

---

### Summary

| Area | Finding | Severity |
|---|---|---|
| `.info` Badge variant — token chain | All tokens correct; dark mode passes WCAG AA | None |
| Portal dropdown — visual parity with `.dropdown` | Confirmed identical visual tokens | None |
| `.addButton` touch target | ~37px computed height < 44px minimum | Medium (non-blocking) |
| `z-index: calc(var(--z-modal) + 100)` | Correct approach; no named token at this level yet | Informational |

The medium finding (touch target) should be fixed before merge or tracked as a refinement item. It does not block this PR given the modal context.

**Verdict: APPROVED** (non-blocking: add `min-height: 44px` to `.addButton` before merge or in refinement)

steilerDev:
**[product-architect]**

Targeted review on architectural compliance, pattern consistency, and ripple effects of the SearchPicker portal change.

**Architecture compliance — PASS**
- Option B respected: no server/shared behavior changes. The diffs under `server/src/services/budgetExtraction/*` and `server/src/services/invoiceAutoItemizeService.test.ts` are pure auto-format reflow (Prettier line-length); no schema, contract, or business-logic drift.
- No new API endpoints, no shared type changes — confidence bucketing (0.85 / 0.6 / 0.3 → `invoice`/`quote`/`professional_estimate`/`own_estimate`) lives entirely in `AutoItemizePage.handleCreateNewBudgetLine`. Acceptable as page-local heuristic; if reused elsewhere later, extract to `client/src/lib/`.

**Portal pattern consistency — PASS**
- `SearchPicker` portal matches the established `WorkItemSelector` pattern: `createPortal(..., document.body)`, `position: fixed`, `getBoundingClientRect()` anchoring, scroll/resize listeners with `capture=true`, click-outside via `[data-search-picker-dropdown]` attribute, Escape handler.
- z-index uses `calc(var(--z-modal) + 100)` = 1100, matching WorkItemSelector's hardcoded 1100. Token-based is the right call.
- Nice addition: viewport flip-above logic when `spaceBelow < 308px` — better than WorkItemSelector's always-below.

**Cross-cutting impact — PASS**
- Verified all SearchPicker consumers: `WorkItemPicker`, `ParentPicker`, `TradePicker`, `AreaPicker`, `HouseholdItemPicker`, `EntityFilter` (DataTable column filters), `MassMoveModal`, `PhotoMetadataSidepanel`, `InvoiceBudgetLinesSection`, `DependencySentenceBuilder` (via WorkItemPicker).
- E2E POM rescoping: `BudgetSourcesPage.ts` (MassMoveModal area) and `InvoiceDetailPage.ts` (WorkItemPicker) correctly switched from modal-scoped to page-scoped `getByRole('option')`. All other E2E references (`work-item-breadcrumb.spec.ts`, `work-item-embeds-breadcrumb.spec.ts`) already use page-scoped `page.getByRole('listbox')` — unaffected.
- `EntityFilter` (DataTable column filter popovers) is an unambiguous improvement — no longer clipped by popover overflow. No E2E uses modal-scoped option lookups for it.
- PhotoMetadataSidepanel uses SearchPicker but has no E2E exercising the dropdown — no fix required.

**Minor observations (non-blocking)**
- `data-search-picker-dropdown` selector inside `handleClickOutside` uses `document.querySelector`, so if two SearchPickers are open simultaneously only the first match is checked. Real-world risk is near-zero (one dropdown open at a time), but consider scoping by ref in a future polish.
- The text input lacks `aria-haspopup="listbox"` / `aria-expanded` / `aria-controls` linking it to the portal listbox. Pre-existing gap (WorkItemSelector has the same), not introduced here.
- `key={opt.id}` for `specialOption` buttons could collide with a real result of the same id — risk negligible given sentinel ids like `THIS_WORK_ITEM`.

None of these block merge. Document them as follow-up polish if you like.

**Verdict: APPROVED**

Issue #1451, opened 2026-05-17
Dark mode CSS variables not reset for print: :global(@media print) in CSS Module is silently dropped by bundler


steilerDev:
# BUG: Dark mode CSS variables not reset for print — :global(@media print) in CSS Module dropped by bundler

**Severity**: Major
**Component**: Frontend UI — Budget Overview print styling (BudgetOverviewPage.module.css)
**Found in**: e2e/tests/budget/budget-overview-print.spec.ts — "Dark mode: print resets CSS variables to light values"

## Steps to Reproduce

1. Navigate to Budget Overview (/budget/overview) with dark mode active (data-theme='dark')
2. Open browser print dialog (or call `window.dispatchEvent(new Event('beforeprint'))` + `page.emulateMedia({ media: 'print' })`)
3. Observe: the page background remains dark

## Expected Behavior

When printing from dark mode, the CSS variables (e.g. `--color-bg-primary`) should be reset to their light-mode equivalents (#ffffff) so the printed output has a white background. This was the intent of the `:global(@media print) { :root { ... } }` block in `BudgetOverviewPage.module.css` (lines 235-251).

## Actual Behavior

The CSS variable reset does NOT apply. In print mode, `getComputedStyle(document.documentElement).getPropertyValue('--color-bg-primary')` returns the dark-mode value instead of `#ffffff`.

## Root Cause (diagnosed)

`BudgetOverviewPage.module.css` contains:

```css
:global(@media print) {
  :root,
  :root[data-theme='dark'] {
    --color-bg-primary: #ffffff;
    /* ... */
  }
}
```

The `:global(@media print)` syntax is **not valid CSS Modules syntax**. CSS Modules supports `:global(.className)` to escape a single class selector, and `:global { ... }` as a block to escape multiple rules — but `:global(@media print)` with an at-rule is not supported and is silently dropped by PostCSS CSS Modules.

**Evidence**: Inspecting the compiled Webpack output (`client/dist/main.*.css`) shows only two `@media print` rules, neither of which contains any `--color-bg-primary` assignment. The variable reset is completely absent from the bundle.

## Fix Suggestion

Move the CSS variable reset to a non-module CSS file (e.g., `client/src/styles/print.css`) or use the correct `:global {}` block syntax:

```css
/* In a .css file (not .module.css), or inside a :global {} block: */
@media print {
  :root,
  :root[data-theme='dark'] {
    --color-bg-primary: #ffffff;
    /* ... */
  }
}
```

[...]

Issue #1450, opened 2026-05-17
usePrintExpansion hook loses pre-print state snapshot due to effect re-run on expandedKeys change


steilerDev:
# BUG-{number}: usePrintExpansion hook does not restore pre-print expansion state after afterprint

**Severity**: Major
**Component**: Frontend UI — Budget Overview print (CostBreakdownTable / usePrintExpansion)
**Found in**: e2e/tests/budget/budget-overview-print.spec.ts — "On-screen expansion state restored after afterprint"

## Steps to Reproduce

1. Navigate to Budget Overview (/budget/overview)
2. Expand Work Items section and the Rohbau area (so Keller area row is visible)
3. Leave the Keller area collapsed (so Kellerbau work item is NOT visible)
4. Open the browser print dialog (or dispatch window.dispatchEvent(new Event('beforeprint')))
5. Observe: all rows become fully expanded (including Kellerbau) — this is correct
6. Close/cancel the print dialog (or dispatch window.dispatchEvent(new Event('afterprint')))
7. Observe: Kellerbau work item remains VISIBLE — it should have been hidden again (pre-print state: Keller collapsed)

## Expected Behavior

After `afterprint` fires, the CostBreakdownTable should restore to the exact expansion state that existed just before `beforeprint` fired. In this case: Work Items expanded, Rohbau expanded, Keller collapsed — so Kellerbau should NOT be visible.

## Actual Behavior

Kellerbau (and all rows that were expanded during print) remain visible after `afterprint`. The table never collapses back to the pre-print state.

## Root Cause (diagnosed)

`usePrintExpansion` captures the pre-print snapshot in a local variable inside its `useEffect`:

```typescript
useEffect(() => {
  let snapshot: Set<string> | null = null;

  function handleBeforePrint() {
    snapshot = new Set(expandedKeys); // ← captured here
    forceExpand();                    // ← setExpandedKeys(allKeys) → triggers React re-render
  }

  function handleAfterPrint() {
    if (snapshot !== null) {
      setExpandedKeys(snapshot);      // ← this handler is REPLACED before afterprint fires
      snapshot = null;
    }
  }

  window.addEventListener('beforeprint', handleBeforePrint);
  window.addEventListener('afterprint', handleAfterPrint);

  return () => { /* cleanup */ };
}, [expandedKeys, forceExpand, setExpandedKeys]); // ← 'expandedKeys' in deps
```

[...]

PR #1606, opened 2026-05-29
fix(budget-print): restore pre-print expansion state and dark-mode print CSS reset (#1450, #1451)

steilerDev:
## Summary

- **#1450 — Expansion state lost after print**: `usePrintExpansion` stored the before-print snapshot in a local `let` variable that was discarded on every effect re-installation. Replaced with `useRef` so the snapshot persists across re-renders and is correctly restored when `window.print()` returns.
- **#1451 — Dark-mode CSS variables not reset during print**: The `:global(@media print)` block was placed inside `BudgetOverviewPage.module.css`, which is invalid CSS Modules syntax (local-scope files cannot use `:global` wrappers around at-rules). Moved the entire CSS variable reset block to the existing global `print.css` file so dark-mode tokens are properly overridden during printing.

Fixes #1450
Fixes #1451

## Test plan

- [ ] Unit tests for `usePrintExpansion`: snapshot is stored via `useRef`, restored after `afterprint`, handles null ref — covered by `usePrintExpansion.test.ts`
- [ ] E2E tests: budget print flow triggers expansion, restores state, and applies correct CSS — covered by `budget-overview-print.spec.ts`
- [ ] Quality Gates CI check passes


steilerDev:
**[product-architect]** APPROVE

Architecture review of the two-bug print fix — all checks pass.

**#1450 (snapshot lost across effect re-run):** Moving the snapshot from a closure-local `let` to `useRef<Set<string>|null>` is the idiomatic React pattern for state that must persist across effect re-runs without triggering re-renders. The closure-local variable was reinitialized to `null` every time `expandedKeys`/`forceExpand` changed (which `beforeprint` itself triggers via `forceExpand`), so `afterprint` could never restore. The ref breaks that coupling correctly. No anti-pattern introduced. The +1 regression test reproduces the exact scenario (snapshot → props change → effect re-run → afterprint restore) and asserts the right outcome.

**#1451 (print CSS-variable reset dropped):** Confirmed `:global(@media print)` is the actual root cause — the CSS Modules PostCSS pipeline does not handle `:global()` wrapping an at-rule and silently drops it. Relocating the block to `client/src/styles/print.css` is the correct architectural home, not a workaround: these resets target document-global selectors (`:root`, `:root[data-theme=dark]`, `html`, `body`, `main`, `#root`) plus `@page` margin boxes — all inherently global concerns that should never live in a page-scoped CSS Module. `print.css` is imported globally in `main.tsx`, alongside the existing print chrome rules, so this is consistent with the established pattern. The `stylelint-disable color-no-hex` justification (anchoring to Layer 1 palette constants) carries over correctly.

**Coverage:** Hook at 100%; the two previously-skipped E2E tests are un-skipped and now exercise the shipped fixes.

No architectural debt, no contract/schema impact. Approving.

steilerDev:
**[ux-designer]**

Design review — `fix/1450-1451-budget-print-fixes` (PR #1606).

**Verdict: APPROVE**

### Token verification — print override block

All 13 hex literals in the new `@media print` block in `print.css` were verified against `client/src/styles/tokens.css`. Every value is the exact resolved Light-mode Layer 1 palette constant for its semantic token:

| Token overridden | Hex in print block | Light-mode palette source | Match |
|---|---|---|---|
| `--color-bg-primary` | `#ffffff` | `--color-white` | ✓ |
| `--color-bg-secondary` | `#f9fafb` | `--color-gray-50` | ✓ |
| `--color-bg-tertiary` | `#f3f4f6` | `--color-gray-100` | ✓ |
| `--color-bg-hover` | `#f9fafb` | `--color-gray-50` | ✓ |
| `--color-text-primary` | `#111827` | `--color-gray-900` | ✓ |
| `--color-text-secondary` | `#374151` | `--color-gray-700` | ✓ |
| `--color-text-body` | `#374151` | `--color-gray-700` | ✓ |
| `--color-text-muted` | `#6b7280` | `--color-gray-500` | ✓ |
| `--color-border` | `#e5e7eb` | `--color-gray-200` | ✓ |
| `--color-border-strong` | `#d1d5db` | `--color-gray-300` | ✓ |
| `--color-success-text-on-light` | `#166534` | `--color-green-700` | ✓ |
| `--color-danger-text-on-light` | `#991b1b` | `--color-red-700` | ✓ |
| `--color-warning-text-on-light` | `#c2410c` | `--color-orange-700` | ✓ |

### Token coverage — dark-mode tokens overridden

The 13-token set correctly covers the semantic tokens that actually flip in dark mode and are relevant to a budget overview printout (backgrounds, body text, muted text, borders, and the three "text-on-light" semantic variants used in budget status cells). The remaining dark-mode overrides (status badges, Gantt, Diary, Budget bar, etc.) are either rendered as SVG/canvas elements that print faithfully at their resolved values, or are not present on the budget overview page.

### Use of hex literals in print.css

Acceptable. The purpose of this block is specifically to override the CSS custom property values themselves — `var()` references would resolve circularly and cannot accomplish this. The file is `print.css`, a global non-module stylesheet that already uses hex literals for print-specific rules elsewhere. The `/* stylelint-disable color-no-hex */` suppression comment is scoped correctly and documents the intent.

### On-screen / dark-mode regression

None. The entire block is inside `@media print`, which browsers only evaluate during print/print-preview. On-screen rendering in both light and dark mode is unaffected.

### Move from CSS Module to global stylesheet

Correct fix. `:global(@media print)` inside a CSS Module is silently dropped by PostCSS/Webpack (the `:global()` escape only works for class selectors, not `@media` blocks targeting `:root`). Moving to `print.css` ensures the rule is actually emitted in the compiled bundle.

Issue #1611, opened 2026-05-29
BUG: "Create Budget Line" in invoice picker does not link new line to invoice (eagerLinkInvoice regression)


steilerDev:
# BUG: \"Create Budget Line\" in invoice picker does not link new line to invoice

**Severity**: Critical
**Component**: Frontend UI / Budget Lines
**Found in**: E2E tests `invoice-budget-line-create-and-link.spec.ts` Scenarios 1, 2, 3

## Steps to Reproduce

1. Navigate to an Invoice Detail page
2. Click \"+ Add Budget Line\"
3. The budget line picker modal opens (step 1)
4. Select a Work Item
5. In step 2, click \"Create Budget Line\"
6. Fill in the BudgetLineForm (description, amount, etc.)
7. Click \"Add Line\" (submit)

## Expected Behavior

A new budget line is created on the work item AND immediately linked to the invoice (appearing in the invoice's Budget Lines section with the planned amount as itemized amount). The picker modal closes and the invoice's remaining amount decreases.

## Actual Behavior

A new budget line is created on the work item, but it is **NOT linked to the invoice**. The picker closes, but the invoice Budget Lines section still shows no lines (or the new line is missing). The invoice remaining amount does not decrease.

## Root Cause

`InvoiceBudgetLinesSection.tsx` (commit `c7d28d71`, PR #1566) sets `eagerLinkInvoice: false` in its `useBudgetLinePicker` call. This causes `handleCreateBudgetLine` in `useBudgetLinePicker` to skip the `createInvoiceBudgetLine` call entirely:

```ts
// useBudgetLinePicker.ts line 302
if (eagerLinkInvoice) {
  // createInvoiceBudgetLine(...) — this is now SKIPPED
}
onLineCreated(newBudgetLine, junctionId); // no-op in InvoiceBudgetLinesSection
closePicker(); // closes immediately — user never gets to link
```

With `eagerLinkInvoice: false` and `onLineCreated: () => {}` (no-op), after creating the budget line via the form, the picker just closes without linking the new line to the invoice.

The component comment says \"No-op: InvoiceBudgetLinesSection handles linking via handleAddSelectedLines\" — but `handleAddSelectedLines` is never called in the create-new flow (only in the existing-line selection flow).

## Environment

- Branch: `beta` (introduced in PR #1566, commit `c7d28d71`, merged 2026-05-25)
- Viewport: all
- Docker: yes

## Evidence

[...]

PR #1617, opened 2026-05-29
fix(invoices): restore eagerLinkInvoice=true in InvoiceBudgetLinesSection (#1611)

steilerDev:
## Summary

Fixes a regression where creating a new budget line through the invoice picker's "Create Budget Line" form no longer linked the line to the invoice.

## Root Cause

`InvoiceBudgetLinesSection` called `useBudgetLinePicker` with `eagerLinkInvoice: false`, which causes `handleCreateBudgetLine` inside the hook to skip the `createInvoiceBudgetLine` call. The `onLineCreated` callback was also a no-op, meaning the newly created budget line simply disappeared after the picker closed — never linked to the invoice.

The confusion arose because `handleAddSelectedLines` (the checkbox multi-select path for *existing* lines) handles linking itself. But the create-new form path depends entirely on the hook's internal eager-link logic.

### `InvoiceBudgetLinesSection.tsx`
- Move `loadBudgetLines` `useCallback` **before** the `useBudgetLinePicker` call so `onLineCreated` can reference it cleanly
- Remove `eagerLinkInvoice: false` — the default (`true`) is now used
- Replace no-op `onLineCreated: () => {}` with `() => { void loadBudgetLines(); }` to refresh the section after create+link

### `InvoiceBudgetLinesSection.test.tsx`
- Update the "happy path" unit test (previously asserted `createInvoiceBudgetLine` was NOT called — documenting the broken behaviour) to verify it IS called with the correct `workItemBudgetId` and `itemizedAmount`

## Testing

- Unit test updated to reflect correct eager-link behaviour
- E2E tests `invoice-budget-line-create-and-link.spec.ts` Scenarios 1/2/3 are unchanged (they encoded correct expected behaviour throughout)

Closes #1611

### DIFF
--- a/client/src/pages/InvoiceDetailPage/InvoiceBudgetLinesSection.test.tsx
+++ b/client/src/pages/InvoiceDetailPage/InvoiceBudgetLinesSection.test.tsx
@@ -882,9 +882,17 @@ describe('InvoiceBudgetLinesSection', () => {
       expect(screen.getByTestId('budget-line-form')).toBeInTheDocument();
     });
 
-    it('submit happy path — direct mode VAT included: calls createWorkItemBudget (not createInvoiceBudgetLine), then closes', async () => {
+    it('submit happy path — creates work item budget line AND eagerly links to invoice (#1611 fix)', async () => {
       const newBudgetLineStub = makeBudgetLineStub('wib-new-001', 500);
       mockCreateWorkItemBudget.mockResolvedValue(newBudgetLineStub);
+
+      // #1611 fix: eagerLinkInvoice is now true so createInvoiceBudgetLine MUST be called
+      const linkedLine = makeDetailLine('ibl-eager-001', {
+        workItemBudgetId: 'wib-new-001',
+        itemizedAmount: 500,
+        plannedAmount: 500,
+      });
+      mockCreateInvoiceBudgetLine.mockResolvedValue(makeCreateResponse(linkedLine, 1000.0));
 
       await openCreateFormWorkItemEmpty();
 
@@ -906,8 +914,16 @@ describe('InvoiceBudgetLinesSection', () => {
         );
       });
 
-      // createInvoiceBudgetLine is NOT called from the create flow (non-eager mode)
-      expect(mockCreateInvoiceBudgetLine).not.toHaveBeenCalled();
+      // createInvoiceBudgetLine IS called immediately after creation (eagerLinkInvoice = true)
+      await waitFor(() => {
+        expect(mockCreateInvoiceBudgetLine).toHaveBeenCalledWith(
+          INVOICE_ID,
+          expect.objectContaining({
+            workItemBudgetId: 'wib-new-001',
+            itemizedAmount: 500,
+          }),
+        );
+      });
 
       // Picker closes after success
       await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeInTheDocument());
--- a/client/src/pages/InvoiceDetailPage/InvoiceBudgetLinesSection.tsx
+++ b/client/src/pages/InvoiceDetailPage/InvoiceBudgetLinesSection.tsx
@@ -92,15 +92,7 @@ export function InvoiceBudgetLinesSection({
   const [selectedLineIds, setSelectedLineIds] = useState<Set<string>>(() => new Set());
   const [itemizedAmounts, setItemizedAmounts] = useState<Record<string, number>>({});
 
-  // Use the picker hook with a no-op callback (InvoiceBudgetLinesSection uses multi-select, not eager link)
-  const picker = useBudgetLinePicker({
-    invoiceId,
-    invoiceAmount: invoiceTotal,
-    onLineCreated: () => {}, // No-op: InvoiceBudgetLinesSection handles linking via handleAddSelectedLines
-    eagerLinkInvoice: false, // Don't auto-link in the create flow
-  });
-
-  // Load budget lines on mount
+  // Load budget lines on mount (defined before picker so onLineCreated can reference it)
   const loadBudgetLines = useCallback(async () => {
     setIsLoading(true);
     setError(null);
@@ -118,6 +110,17 @@ export function InvoiceBudgetLinesSection({
       setIsLoading(false);
     }
   }, [invoiceId, t]);
+
+  // Use the picker hook — eagerLinkInvoice defaults to true so the create-new flow
+  // calls createInvoiceBudgetLine automatically before invoking onLineCreated.
+  const picker = useBudgetLinePicker({
+    invoiceId,
+    invoiceAmount: invoiceTotal,
+    onLineCreated: () => {
+      // Reload budget lines after a new line is created and eagerly linked to the invoice
+      void loadBudgetLines();
+    },
+  });
 
   useEffect(() => {
     void loadBudgetLines();


steilerDev (commented):
**[product-owner]**

## Product Owner Review — PR #1617

Reviewed against Issue #1611 acceptance criteria.

---

### Issue #1611 — Acceptance Criteria Verification

**AC1 — Create Budget Line links to invoice automatically**
✅ **Met.** Removing `eagerLinkInvoice: false` (restoring the default `true`) causes `handleCreateBudgetLine` inside `useBudgetLinePicker` to call `createInvoiceBudgetLine` immediately after the work-item budget line is created. The hook's internal eager-link path is now exercised, which is exactly what was skipped by the regression.

**AC2 — New budget line appears in the invoice's Budget Lines section after submission**
✅ **Met.** `onLineCreated` is now wired to `void loadBudgetLines()` instead of a no-op, so the section refreshes its data from the server after the create+link completes. The Budget Lines section will reflect the new line with the correct amount.

**AC3 — Picker modal closes after successful submission**
✅ **Met.** The diff confirms the picker-close assertion (`expect(screen.queryByRole('dialog')).not.toBeInTheDocument()`) is retained and still passing in the updated test. The hook's `closePicker` call path is unchanged.

**AC4 — Invoice remaining amount decreases (eager link records itemizedAmount)**
✅ **Met.** The updated test asserts `createInvoiceBudgetLine` is called with `workItemBudgetId: 'wib-new-001'` and `itemizedAmount: 500`, confirming the correct payload is sent to the server so the invoice total is updated accurately.

**AC5 — E2E tests (Scenarios 1, 2, 3) pass without modification**
✅ **Met.** PR description explicitly states the E2E tests `invoice-budget-line-create-and-link.spec.ts` are unchanged — they encoded the correct expected behaviour throughout and should now pass.

**AC6 — Existing multi-select (checkbox) path for linking existing lines is unaffected**
✅ **Met.** The `handleAddSelectedLines` code path is untouched by this diff. The change is scoped entirely to the `useBudgetLinePicker` instantiation and the `loadBudgetLines` reordering, with no changes to the selection or linking logic for existing budget lines.

---

### Root Cause Confirmation

The regression diagnosis in #1611 is precisely correct and the fix directly addresses it: `eagerLinkInvoice: false` + a no-op `onLineCreated` meant the create-form path created the work-item budget line but never linked it to the invoice and silently closed. Restoring the default (`true`) and wiring the callback to reload closes both gaps in one minimal, targeted change.

### Code Quality Observations

- Reordering `loadBudgetLines` before the `useBudgetLinePicker` call is the right structural fix — it eliminates a closure dependency concern and the comment explains the intent clearly.
- The unit test title is updated to reflect the *correct* behaviour rather than carrying forward a comment that documented broken behaviour — good practice.
- The change is appropriately narrow: two files, no API contract changes, no side-effects on other components.

---

### Verdict: ✅ APPROVED (self-review restriction prevented formal approval)

All acceptance criteria from #1611 are satisfied. The fix is correct, minimal, and well-tested. Safe to merge.

*Co-Authored-By: Claude product-owner (Opus 4.6) <noreply@anthropic.com>*

steilerDev (commented):
**[ux-designer]**

## UX Design Review — PR #1617 `fix(invoices): restore eagerLinkInvoice=true`

**Verdict: ✅ Approve** — This is a pure behavioural bug fix with zero visual/layout/token changes. All design system criteria are met.

---

### Scope Confirmation

Only two files are modified:

| File | Nature of Change |
|------|-----------------|
| `InvoiceBudgetLinesSection.tsx` | Logic-only: declaration order of `loadBudgetLines` reordered; `eagerLinkInvoice: false` removed; `onLineCreated` given a real body |
| `InvoiceBudgetLinesSection.test.tsx` | Test updated to assert correct eager-link behaviour |

No `.module.css`, `tokens.css`, `shared.module.css`, or any other style file is touched. The design system checklist items below are all unaffected/compliant by inspection.

---

### Design Review Checklist

- [x] **Token adherence** — No hardcoded colors, sizes, or spacing introduced. No CSS changes at all.
- [x] **Visual consistency** — No new UI elements added; existing component structure unchanged.
- [x] **Dark mode correctness** — No color values modified; dark mode behaviour is unaffected.
- [x] **Responsive implementation** — No layout changes; responsive behaviour is unaffected.
- [x] **Accessibility** — No ARIA roles, keyboard handlers, or focus management removed. The existing focus-return to `newLineRowRef` after picker close remains intact. The `aria-live="polite"` region on the remaining-amount cell continues to announce updates correctly.
- [x] **Shared pattern usage** — `loadBudgetLines` was already using `useCallback`; no new patterns introduced.
- [x] **Component reuse** — No duplication of Badge, SearchPicker, Modal, Skeleton, EmptyState, or FormError.
- [x] **Animation/transition** — No animation or transition code modified.

---

### UX Flow Verification ✅

The requested flow was verified against the post-fix source:

1. **User opens picker → selects a work/household item → clicks "Create Budget Line"**
   The picker hook (`useBudgetLinePicker`) now runs with `eagerLinkInvoice` defaulting to `true`, so `handleCreateBudgetLine` inside the hook calls `createInvoiceBudgetLine` immediately after `createWorkItemBudget`/`createHouseholdItemBudget` succeeds.

2. **Picker closes automatically** — `closePicker()` is invoked by the hook after the eager link, and focus returns to `newLineRowRef` (the most-recently-added row) within 100ms — the correct focus destination per the existing focus management design.

3. **Section refreshes showing the new linked line** — `onLineCreated` now calls `void loadBudgetLines()`, which re-fetches the full `InvoiceBudgetLineDetailResponse` array from the server, populating both `budgetLines` state and `remainingAmount`. The `aria-live="polite"` cell on the remaining amount will announce the updated value to screen readers without interrupting the user.

4. **Remaining amount updates correctly** — Because `loadBudgetLines` fetches the authoritative `remainingAmount` from the API response rather than computing it client-side, the displayed value is always consistent with server state — including any edge cases where the link response includes rounding adjustments.

**The flow is correct, coherent, and accessible as implemented.**

---

### One Informational Note (non-blocking)

The `loadBudgetLines` full-reload approach is intentionally conservative (vs. optimistically prepending the new line from the link response, as the multi-select path does in `handleAddSelectedLines`). This is a good defensive choice for the create path since the server may normalise the `itemizedAmount` or attach additional metadata. No design concern here.

---

*Reviewed by `ux-designer` · Co-Authored-By: Claude ux-designer (Sonnet 4.6) <noreply@anthropic.com>*

steilerDev (commented):
**[product-architect]**

## Review: fix(invoices): restore eagerLinkInvoice=true in InvoiceBudgetLinesSection (#1611)

**Verdict: ✅ Approve** *(posted as comment — cannot self-approve)*

This is a clean, well-scoped bug fix. I reviewed the full diff against the beta base, the `useBudgetLinePicker` hook implementation, the hook's own test suite, and the component test file. Everything checks out.

---

### Root Cause Confirmed

The regression on `beta` was clear from the diff. The base had:

```ts
const picker = useBudgetLinePicker({
  invoiceId,
  invoiceAmount: invoiceTotal,
  onLineCreated: () => {},       // no-op — new line silently dropped
  eagerLinkInvoice: false,       // skip createInvoiceBudgetLine entirely
});

// loadBudgetLines defined AFTER picker — so onLineCreated couldn't reference it
const loadBudgetLines = useCallback(...);
```

The `handleCreateBudgetLine` path inside the hook gates the `createInvoiceBudgetLine` call behind `eagerLinkInvoice`. With it `false`, the newly created budget line was never linked to the invoice, and the no-op `onLineCreated` meant the section never refreshed. The bug was complete and silent.

---

### Hook Ordering — Correct ✅

Moving `loadBudgetLines` (as a `useCallback`) **before** the `useBudgetLinePicker()` call is the right fix. In JavaScript/React, `useCallback` is evaluated immediately on every render, so the `loadBudgetLines` reference passed into `onLineCreated` is always the stable, current callback by the time `useBudgetLinePicker` captures it in its own `useCallback` dependency array (`onLineCreated` is listed there). Hook ordering rules are respected — this is not a conditional hook, just a reorder.

---

### API Call Correctness ✅

The `onLineCreated` callback now calls `void loadBudgetLines()`, which does a fresh `GET /api/invoices/:id/budget-lines` — exactly right. The hook itself handles the optimistic `setBudgetLines`/`setRemainingAmount` update synchronously (via the `createInvoiceBudgetLine` response), but `loadBudgetLines` ensures the section reflects server truth after the picker closes. No double-linked-state risk because the picker closes before `onLineCreated` fires (`closePicker()` is called inside the hook immediately before `onLineCreated()`).

---

### Default Parameter Removal ✅

Removing the explicit `eagerLinkInvoice: false` and relying on the hook's default (`eagerLinkInvoice = true`) is semantically correct and reduces noise. The hook's TypeScript interface documents the default clearly.

---

### Test Changes ✅

The updated test is a faithful description of the **correct** behaviour. The previous assertion (`expect(mockCreateInvoiceBudgetLine).not.toHaveBeenCalled()`) literally encoded the bug. The replacement verifies:
1. `createWorkItemBudget` is called with the right payload
2. `createInvoiceBudgetLine` is called with `workItemBudgetId` and `itemizedAmount` matching the created line
3. The dialog closes after success
4. The newly linked line appears in the table

The `mockCreateInvoiceBudgetLine` setup (providing a `makeCreateResponse`) is also required for the test to reach the close-picker path — without it the hook would throw and the test would fail for the wrong reason. Good hygiene.

---

### No Backend / Schema Changes Needed ✅

Confirmed. The fix is entirely in the client layer. `createInvoiceBudgetLine` and `createWorkItemBudget` API calls were already contracted and implemented — this just re-enables a call path that was incorrectly disabled.

---

### Minor Observations (informational only, not blocking)

1. **`onLineCreated` signature drift**: The hook's `onLineCreated` callback is typed as `(line, junctionId) => void` but the `InvoiceBudgetLinesSection` implementation ignores both parameters and just calls `loadBudgetLines()`. This is fine for correctness — the full reload is the right approach here — but if other consumers need the `junctionId` for optimistic updates they should use the parameters. Not a concern for this PR.

2. **Inline `onLineCreated` lambda re-creates on every render**: The `() => { void loadBudgetLines(); }` arrow is a new reference each render, which causes `handleCreateBudgetLine` inside the hook to re-create (it lists `onLineCreated` in its deps). Cheap and not a correctness problem, but could be addressed with a `useCallback` wrapper at the call site in a follow-up if the exhaustive-deps lint rule fires. Not a blocker — this pattern is already present throughout the codebase.

Both observations are pre-existing patterns, neither introduced by this PR.

Issue #1458, opened 2026-05-17
chore: clean up @typescript-eslint/consistent-type-imports (44 errors)


steilerDev:
**Parent tracking issue**: #1455

## Summary

Resolve all 44 \`@typescript-eslint/consistent-type-imports\` errors.

## Fix pattern

Swap value imports for type-only imports where the imported symbol is only used in type positions:

\`\`\`ts
// before
import { Foo } from './foo.js';
function bar(x: Foo): void {}

// after
import type { Foo } from './foo.js';
function bar(x: Foo): void {}
\`\`\`

Most of these are auto-fixable with \`npm run lint:fix\`. Run that first, then hand-fix anything the auto-fixer can't reason about (mixed type/value imports from the same module).

## Scope

All workspaces. Heavier in \`client/\` and \`server/\`.

## Acceptance Criteria

- [ ] \`npm run lint\` reports 0 \`@typescript-eslint/consistent-type-imports\` errors
- [ ] \`npm run typecheck\` passes
- [ ] No runtime behavior changes (type-only imports are erased at compile time)

## Risk

Very low. Mostly mechanical, mostly auto-fixable.

steilerDev:
**Status update**: Attempted automated fix in PR #1630, but closed after 3 fix-loop iterations. 

**Root cause**: The test files use complex namespace import patterns (`import * as XModule from './X.js'`) that have been refactored in beta to use inline `typeof import(...)` syntax. Automated agents kept either:
- Converting `import * as X` → `import type * as X` (causes TS2709)
- Or removing the import entirely (causes TS2304)

**Recommendation**: Use `npx eslint --fix` in a proper dev environment, or handle this manually. The 44 errors are all of type 'add the `type` keyword to existing named imports' — but the files are complex enough that line-by-line automated edits keep breaking things.

Skipping for now in the batch-develop run and continuing with remaining issues.

Issue #1455, opened 2026-05-17
Tracking: zero out ESLint warnings + errors


steilerDev:
## Context

Baseline lint state on `beta` is **857 problems (409 errors + 448 warnings)**. PR #1452 attempted an all-in-one sweep but was closed after the bundled diff kept introducing TypeScript regressions that were hard to isolate. This tracking issue breaks the cleanup into focused per-rule chores so each one can ship as a small, low-risk PR.

See PR #1452 for the prior attempt and the commits that did land cleanly (notably the eslint-config test-files override in commit `7f8ab3e0`).

### Errors (409)
- 162 `@typescript-eslint/no-unused-vars`
- 119 `@eslint-react/component-hook-factories` (mostly test files)
- 44 `@typescript-eslint/consistent-type-imports`
- 26 `@eslint-react/unsupported-syntax` (IIFE in JSX)
- 10 `@eslint-react/jsx-no-children-prop-with-children`
- 4 `@eslint-react/static-components`
- 4 `@eslint-react/error-boundaries`
- 2 `@eslint-react/rules-of-hooks`
- 2 `@eslint-react/no-nested-component-definitions`
- 2 `@typescript-eslint/no-empty-object-type`

### Warnings (448)
- 222 `@typescript-eslint/no-explicit-any`
- 146 `@eslint-react/no-unnecessary-use-prefix` (mostly tests)
- 76 `@eslint-react/set-state-in-effect`
- 74 `@eslint-react/exhaustive-deps`
- 36 `@eslint-react/use-state` (lazy init)
- 34 `@eslint-react/no-array-index-key`
- 34 `@eslint-react/jsx-no-children-prop`
- 14 `@eslint-react/naming-convention-ref-name`
- 8 `@eslint-react/web-api-no-leaked-timeout`
- 4 `@eslint-react/no-use-context`
- 4 `@eslint-react/no-context-provider`
- 2 `@eslint-react/web-api-no-leaked-fetch`
- 2 `@eslint-react/purity`

## Process

Each child issue should be addressed as its own PR. Do not bundle multiple rule families into one PR — the bundling is what sank PR #1452.

**Suggested ordering**: ship #1456 (test-files config override) first — it drops ~280 warnings without touching code, which makes every subsequent issue smaller and easier to verify.

## Child issues

- [ ] #1456 — chore: relax React Compiler rules in eslint config for test files (ship first)
[...]

PR #1630, opened 2026-05-29
chore(eslint): fix @typescript-eslint/consistent-type-imports (#1458)

steilerDev:
## Summary

Converts value imports to `import type` in all files where the imported symbol is only used in TypeScript type positions (type annotations, interface definitions, generic type parameters).

## Files changed (15 files, all test files)

The production code already uses `import type` correctly. The remaining violations were in test files:

- `client/src/components/` — 11 test files
- `server/src/routes/` — 1 test file  
- `server/src/services/` — 3 test files

## Pattern applied

```ts
// Before
import { SomeType } from './module.js';
// After (type is only used in type positions)
import type { SomeType } from './module.js';
```

## Related
- Parent tracking issue: #1455
- Closes #1458

### DIFF
--- a/client/src/components/MassMoveModal/MassMoveModal.test.tsx
+++ b/client/src/components/MassMoveModal/MassMoveModal.test.tsx
@@ -7,7 +7,6 @@ import type React from 'react';
 import type { BudgetSource, BudgetSourceListResponse } from '@cornerstone/shared';
 import { ApiClientError } from '../../lib/apiClient.js';
 import type { MassMoveModalProps } from './MassMoveModal.js';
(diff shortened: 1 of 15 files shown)


steilerDev:
Closing without merging after 3 fix-loop iterations. The automated approach kept introducing TypeScript errors (TS2709/TS2304) due to complex namespace import patterns (`import * as Module`) in test files. These files use `typeof import(...)` inline patterns that our agents mistakenly converted. Issue #1458 needs manual attention or a more targeted approach (e.g., running `eslint --fix` in a dedicated environment).

PR #1683, opened 2026-06-15
chore(deps): align root and client konva versions to prevent type collision (#1646)

### LINKED ISSUE #1646: chore(deps): align root and client Konva versions to eliminate type-collision
## Problem

Two `konva` installations coexist in `node_modules` with mutually incompatible `Shape<ShapeConfig>` interfaces, causing TypeScript errors any time a workspace dependency change forces a fresh lockfile resolution.

- **Root**: `/node_modules/konva` — older major (9.x). Its `Shape<ShapeConfig>` does **not** declare a `miterLimit` property.
- **Client workspace**: `/client/node_modules/konva` — Konva 10.x (pinned by `client/package.json`). Its `Shape<ShapeConfig>` **requires** `miterLimit`.

Because tsc resolves `import { Stage, Layer, Shape } from 'konva/lib/...'` paths that overlap between root and client workspaces, the two type trees collide and produce errors like:

```
error TS2322: Type 'RefObject<Stage | null>' is not assignable to type 'Ref<Stage> | undefined'.
  Type 'RefObject<.../client/node_modules/konva/.../Stage | null>'
    is not assignable to type
  'RefObject<.../node_modules/konva/.../Stage | null>'.
    Property 'miterLimit' is missing in type
    '.../node_modules/konva/.../Shape<ShapeConfig>'
    but required in type
    '.../client/node_modules/konva/.../Shape<ShapeConfig>'.
```

Reproduced on `client/src/components/photos/PhotoAnnotator/PhotoAnnotator.tsx` lines 1248 and 1252.

## Why it's invisible today

Some transitive dependency in the root workspace pulls in `konva@9.x`. As long as the lockfile is left untouched, `npm` continues to resolve that version. The current beta lockfile happens to be in a stable state where Static Analysis passes (verified: PRs #1640 and #1644 both green this week).

## What triggers it

Any change that **regenerates the root-level `package-lock.json`** can re-resolve the root-level `konva` to a version that drifts from the client workspace's pinned 10.x. Demonstrated by #1620: adding `ts-node@10.9.2` as an explicit root devDep and running `npm install --package-lock-only` was enough to break Static Analysis with the error above. This is a latent landmine for any future root-deps change.

## Proposed fixes (pick one)

**Option 1 — npm overrides (minimal change)**
Add an `overrides` block to root `package.json` pinning `konva` to whatever version the client workspace uses (currently 10.x). Forces npm to hoist a single resolved version for the entire workspace tree.

```json
{
  \"overrides\": {
    \"konva\": \"10.x.x\"
  }
}
```

**Option 2 — explicit root devDep**
Add `konva` as an explicit root devDep at the same version the client workspace pins. Removes the transitive-resolution ambiguity by making npm see one resolver target.

**Option 3 — eliminate root-level konva**
Audit which transitive dep is dragging in `konva@9.x` at the root and either remove it or upgrade it past the konva drop. Cleanest long-term fix; highest investigation cost.

## Acceptance criteria

- A fresh `npm install` (or `npm install --package-lock-only`) starting from a wiped lockfile produces a tree where **only one** version of `konva` exists across root + client workspaces.
- Static Analysis passes after lockfile regeneration without further code changes.
- Add a guardrail (CI check or precommit) to fail when more than one `konva` version is resolved.

## Context

- Surfaced during WP-0 PR janitorial (2026-05-30) while attempting to land deferred PR #1620.
- Until this issue is resolved, **PRs that touch root `package.json` / `package-lock.json` will likely hit Static Analysis failures on rebase against beta.**
- Tracker: #1573 (Jest 30.4.x adoption) remains open; #1620 was closed deferred because its lockfile regen tripped this exact bug.

steilerDev:
## Summary

- Add a root `package.json` `overrides` entry pinning `konva` to `10.3.0` (matching the client workspace pin) so npm hoists a single konva version across the workspace tree
- Prevents a latent 9.x/10.x version split that produces `TS2322 Shape<ShapeConfig>` type-collision errors on lockfile regeneration
- Add `scripts/check-konva-single-version.sh`: a CI guardrail that fails when more than one konva version is resolved in `package-lock.json`, wired into the `static-analysis` job after `npm ci`

Fixes #1646

## Test plan

- [ ] `scripts/check-konva-single-version.sh package-lock.json` exits 0 (single version `10.3.0` resolved)
- [ ] CI static-analysis job runs the new konva check step without error
- [ ] `Quality Gates` CI gate passes

Co-Authored-By: Claude backend-developer (Haiku 4.5) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Check single Konva version
+        run: bash scripts/check-konva-single-version.sh package-lock.json
+
       - name: Verify dependency signatures
         run: npm audit signatures
 
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "picomatch@>=4.0.0 <4.0.4": "4.0.4",
     "react": "19.2.6",
     "react-dom": "19.2.6",
-    "serialize-javascript": "7.0.5"
+    "serialize-javascript": "7.0.5",
+    "konva": "10.3.0"
   }
 }
--- a/scripts/check-konva-single-version.sh
+++ b/scripts/check-konva-single-version.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+# check-konva-single-version.sh
+#
+# Verifies that exactly one version of konva is resolved across the entire
+# npm workspace tree by inspecting package-lock.json.
+#
+# A duplicate konva installation causes mutually incompatible Shape<ShapeConfig>
+# interfaces between 9.x and 10.x, leading to TS2322 type errors.
+#
+# Usage: check-konva-single-version.sh <path-to-package-lock.json>
+#
+# Exits non-zero with a clear error message if more than one distinct konva
+# version is found.
+
+set -euo pipefail
+
+LOCKFILE="${1:-package-lock.json}"
+
+if [[ ! -f "$LOCKFILE" ]]; then
+  echo "check-konva-single-version: lockfile not found: $LOCKFILE" >&2
+  exit 1
+fi
+
+RESULT=$(LOCKFILE="$LOCKFILE" node --input-type=module <<'EOF'
+import { readFileSync } from 'fs';
+
+const lock = JSON.parse(readFileSync(process.env.LOCKFILE, 'utf8'));
+const packages = lock.packages ?? {};
+
+// Collect every resolved konva entry: hoisted (node_modules/konva) and
+// any nested installation (*/node_modules/konva).
+const entries = Object.entries(packages)
+  .filter(([k]) => k === 'node_modules/konva' || k.endsWith('/node_modules/konva'));
+
+const versions = [...new Set(entries.map(([, v]) => v.version))];
+
+if (versions.length === 0) {
+  // konva not installed at all — not this script's concern
+  process.exit(0);
+}
+
+if (versions.length > 1) {
+  process.stdout.write('MULTIPLE:' + versions.join(',') + '\n');
+} else {
+  process.stdout.write('OK:' + versions[0] + '\n');
+}
+EOF
+)
+
+if [[ "$RESULT" == MULTIPLE:* ]]; then
+  VERSIONS="${RESULT#MULTIPLE:}"
+  echo "" >&2
+  echo "ERROR: Multiple konva versions found in ${LOCKFILE}:" >&2
+  echo "  ${VERSIONS//,/  }" >&2
+  echo "" >&2
+  echo "  Incompatible konva versions cause TS2322 type-collision errors" >&2
+  echo "  (Shape<ShapeConfig> interfaces are mutually incompatible between" >&2
+  echo "  the 9.x and 10.x series)." >&2
+  echo "" >&2
+  echo "  Fix: ensure the root package.json 'overrides' block pins konva to" >&2
+  echo "  exactly one version, then run 'npm install' to regenerate the lockfile." >&2
+  echo "" >&2
+  exit 1
+fi
+
+if [[ "$RESULT" == OK:* ]]; then
+  echo "check-konva-single-version: OK (konva ${RESULT#OK:})"
+fi


steilerDev:
**[product-architect]** APPROVE

Reviewed against architecture/dependency-policy conventions. Change set is build/CI/tooling only (no production source). Verified locally against the worktree.

**Architecture choice (Option 1: npm overrides) — correct.** konva is not a direct root dependency; it is pulled transitively (client declares `konva@10.3.0` + `react-konva@19.2.4`). Eliminating a "root konva" is a non-issue (there is none), so the override is the right lever: it forces a single hoisted resolution across the workspace tree regardless of how npm walks the graph. Consistent with the existing overrides block (`react`, `react-dom`, `serialize-javascript`, etc.).

**Dependency policy compliance — passes.** Override value `"konva": "10.3.0"` is exact-pinned (no caret/tilde). Overrides are correctly exempt from the direct-dependency caret rule, matching every other entry in the block.

**Acceptance criterion #1 (wiped-lockfile reinstall → single konva) — genuinely satisfied.** Confirmed `react-konva@19.2.4` declares peer `^8.0.1 || ^7.2.5 || ^9.0.0 || ^10.0.0`. Without the override, a from-scratch reinstall could legally hoist a 9.x build and collide with the client's 10.x, producing the TS2322 `Shape<ShapeConfig>` mismatch the script documents. The override constrains resolution to 10.3.0 across all branches of the tree. Lockfile is correctly left unchanged — it already resolves to a single `node_modules/konva@10.3.0` with no nested copy, so the override is purely forward-looking and no churn is warranted.

**Guardrail correctness — verified all branches:**
- Real lockfile → `OK (konva 10.3.0)`, exit 0
- Missing lockfile → error, exit 1
- konva absent → silent, exit 0 (correct: not the script's concern)
- Two distinct versions → clear error + remediation, exit 1

`set -euo pipefail`, embedded ESM node parse of `package-lock.json`, and the `node_modules/konva` + `*/node_modules/konva` match pattern are all sound and mirror the established `scripts/check-dep-pinning.sh` style. CI placement (after `npm ci`, before `npm audit signatures`, in `static-analysis`) is appropriate — it gates on the freshly-installed lockfile.

**Non-blocking observations (no change required for this PR):**
- The sibling guard `scripts/check-dep-pinning.sh` is not wired into `ci.yml`; this PR newly establishes the pattern of running such guards in CI. Worth a follow-up to also wire in the pinning check for full coverage — out of scope here.
- The guard reads the committed lockfile, not the installed `node_modules` tree, so it validates intent rather than the actual install. That is the correct and deterministic choice for CI; noting it only so future readers understand the guarantee is "lockfile resolves to one version," which combined with `npm ci` (which fails on lockfile drift) is sufficient.

No concerns that block merge to beta.

steilerDev:
**[security-engineer]** PR #1683 — Security Review: Align Konva versions (chore/#1646)

## Summary

This PR makes two changes: adds a `konva: 10.3.0` entry to the root `overrides` block in `package.json` to enforce a single resolved version tree-wide, and introduces a new CI validation script (`scripts/check-konva-single-version.sh`) that runs as part of the `static-analysis` job.

No production source files, API routes, authentication logic, or database code are modified. The risk surface is limited to supply-chain pinning behavior and CI script safety.

---

## Findings

### [INFORMATIONAL] Shell heredoc interpolates LOCKFILE path directly into Node.js source

**OWASP Category**: A03 — Injection (not directly exploitable in this context)
**Severity**: Informational
**Status**: Open

The heredoc in `scripts/check-konva-single-version.sh` (line 27) interpolates the `LOCKFILE` shell variable directly into the Node.js source string passed to `node --input-type=module`:

```
const lock = JSON.parse(readFileSync('${LOCKFILE}', 'utf8'));
```

If `LOCKFILE` were ever set to a path containing a single-quote or path-traversal characters (e.g., `'); process.exit(0); //`), the interpolation would inject arbitrary Node.js code into the inline script before it is executed.

**In the current CI invocation** (`bash scripts/check-konva-single-version.sh package-lock.json`), the argument is a hardcoded repository-controlled literal, so there is no attack vector. This would only matter if the script were ever called with untrusted input (e.g., from a pull-request-triggered workflow that constructs the lockfile path from PR metadata). In the current workflow structure, this is safe.

**Recommendation** (optional hardening — not blocking): Pass the path to Node.js as an environment variable instead of interpolating into source code. Changing the heredoc delimiter to `'EOF'` (single-quoted) prevents shell expansion inside the heredoc, but then requires the path to be passed through the environment:

```bash
# Use single-quoted 'EOF' to stop shell expansion inside the heredoc,
# then pass the path via an environment variable
RESULT=$(LOCKFILE="$LOCKFILE" node --input-type=module <<'JSEOF'
import { readFileSync } from 'fs';
const lock = JSON.parse(readFileSync(process.env.LOCKFILE, 'utf8'));
// ... rest of script
JSEOF
)
```

---

## Checklist

- [x] No SQL/command/XSS injection vectors in new code (heredoc interpolation is theoretical — CI arg is hardcoded)
- [x] No new endpoints; auth/authorization unaffected
- [x] No sensitive data in logs, errors, or responses
- [x] No new dependencies introduced; konva 10.3.0 is a version already resolved in the tree
- [x] CI step runs no network calls, reads only a local checked-in file, adds no new workflow permissions

---

## Verdict

**APPROVE.** No blocking findings. The `package.json` override pins konva to the already-resolved 10.3.0 — no new package is introduced. The CI script is correct, runs with a fixed trusted argument, and does not introduce a real injection risk in its current invocation context. The single informational note above is a defense-in-depth suggestion for future callers of the script.

steilerDev:
Closing as **superseded by #1685**.

While this PR was in flight, #1685 merged to \`beta\` and resolved #1646 with the same \`konva@10.3.0\` \`overrides\` pin plus a *more general* guardrail — \`scripts/check-single-dep-version.sh <pkg>\` (npm-ls based, reusable across packages), wired into the \`static-analysis\` CI job as \`check-single-dep-version.sh konva\`.

This PR's konva-specific \`scripts/check-konva-single-version.sh\` offers no advantage over beta's generic checker, and its \`package.json\` override + CI step now duplicate/conflict with beta. Issue #1646 is already closed. No action needed here.

PR #1685, opened 2026-06-15
chore(deps): align root and client Konva versions to eliminate type-collision (#1646)

### LINKED ISSUE #1646: chore(deps): align root and client Konva versions to eliminate type-collision

steilerDev:
## Summary

- Add `konva` (exact pin `10.3.0`) to the root `package.json` `overrides` block so the entire npm workspace tree hoists a single Konva version — eliminating the latent root-vs-client type collision (`Shape<ShapeConfig>` `miterLimit` mismatch) that surfaces whenever the root lockfile is regenerated.
- Add `scripts/check-single-dep-version.sh` — a guardrail that fails when more than one version of a named package resolves across the workspace tree.
- Wire a `Check single konva version` step into the CI `static-analysis` job (after `Install dependencies`) so a future regression is caught with a clear error instead of an opaque TS2322.

The lockfile intentionally has no diff: Konva already resolves to `10.3.0` everywhere today, so the override is a forward-looking constraint that guarantees a single version after any future lockfile regeneration (AC1).

Fixes #1646

## Test plan

- [x] Guardrail validated: single-version pass (exit 0), simulated multi-version fail (exit 1 + error block), no false positive on a missing package (exit 0)
- [x] CI step positioned after `npm ci`, before typecheck/build
- [ ] Static Analysis (typecheck + build) green on CI

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Check single konva version
+        run: bash scripts/check-single-dep-version.sh konva
+
       - name: Verify dependency signatures
         run: npm audit signatures
 
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
   "overrides": {
     "minimatch@>=10.0.0 <10.2.3": "10.2.4",
     "picomatch@>=4.0.0 <4.0.4": "4.0.4",
+    "konva": "10.3.0",
     "react": "19.2.6",
     "react-dom": "19.2.6",
     "serialize-javascript": "7.0.5"
--- a/scripts/check-single-dep-version.sh
+++ b/scripts/check-single-dep-version.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+# check-single-dep-version.sh
+#
+# Verifies that exactly one version of each named package resolves across the
+# entire workspace dependency tree. A second copy of a package (often pulled in
+# transitively at a different major/minor) can cause duplicate type definitions
+# and runtime type-collision bugs (e.g. konva surfacing two incompatible sets of
+# .d.ts files).
+#
+# Usage: check-single-dep-version.sh <package-name> [<package-name> ...]
+#
+# For each package:
+#   - exits 0 with a confirmation line when exactly one version (or zero, for a
+#     package that does not resolve at all) is present
+#   - records a failure and prints an error block (the resolved versions plus a
+#     remediation hint) when more than one version resolves
+#
+# Exits 1 if any named package resolves to more than one version.
+
+set -euo pipefail
+
+FAILED=0
+
+for PKG in "$@"; do
+  # Collect the dependency tree as JSON. `npm ls <pkg> --all --json` emits the
+  # full tree; a small node walker flattens it and prints the unique resolved
+  # versions, one per line. The JSON is written to a temp file (the tree can be
+  # large, exceeding argv/env size limits) and its path plus the package name
+  # are passed to node as arguments, leaving stdin free for the script body.
+  LS_FILE=$(mktemp)
+  npm ls "$PKG" --all --json >"$LS_FILE" 2>/dev/null || true
+
+  VERSIONS=$(node --input-type=module - "$PKG" "$LS_FILE" <<'EOF'
+import { readFileSync } from 'fs';
+
+const target = process.argv[2];
+const lsFile = process.argv[3];
+
+let tree;
+try {
+  tree = JSON.parse(readFileSync(lsFile, 'utf8'));
+} catch {
+  // No parseable output (e.g. empty) => no versions resolved.
+  process.exit(0);
+}
+
+const versions = new Set();
+
+const walk = (node) => {
+  if (!node || typeof node !== 'object') return;
+  const deps = node.dependencies;
+  if (!deps) return;
+  for (const [name, child] of Object.entries(deps)) {
+    if (name === target && child && typeof child.version === 'string') {
+      versions.add(child.version);
+    }
+    walk(child);
+  }
+};
+
+walk(tree);
+
+for (const v of versions) {
+  process.stdout.write(v + '\n');
+}
+EOF
+)
+
+  rm -f "$LS_FILE"
+
+  # Count distinct versions (filter out blank lines).
+  COUNT=$(printf '%s\n' "$VERSIONS" | grep -c . || true)
+
+  if [[ "$COUNT" -le 1 ]]; then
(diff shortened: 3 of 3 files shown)


steilerDev:
**[security-engineer]** APPROVED — no findings

Reviewed three angles on this dependency-alignment chore:

**1. konva 10.3.0 version safety**
The override pins konva to 10.3.0, which is the same exact version already declared in `client/package.json`. No version change occurs — this only prevents a second copy of konva at a different version from sneaking in transitively. `npm audit --omit=dev` is clean; no known CVEs in konva 10.3.0.

**2. Shell script security (`scripts/check-single-dep-version.sh`)**
- `set -euo pipefail` is present — unset variables and pipeline failures are caught.
- `mktemp` creates a unique temp file; the path is passed to `node` as an argument (not interpolated into a shell command string), so there is no temp-file race or injection vector.
- The only external input is `$PKG` from `"$@"`. In the CI invocation this is the hard-coded string `konva` — no user-supplied data. Even in a hypothetical manual call with an adversarial package name, it is passed to `npm ls` and as an argv element to node; it is never interpolated into an `eval` or unquoted shell expansion.
- The inline node script reads the temp file with `readFileSync` and parses it with `JSON.parse` — no `eval`, no `vm.runInNewContext`, no deserialisation risk.
- `rm -f "$LS_FILE"` is called after the node script completes; the temp file is cleaned up on both success and failure paths (it executes unconditionally before the `COUNT` check).

**3. CI workflow step**
The new step runs `bash scripts/check-single-dep-version.sh konva` — a repo-local script with a hard-coded argument. It does not fetch external tooling, escalate permissions, or access secrets. The `static-analysis` job already has `contents: read` scope (the workflow's default). No supply-chain or privilege-escalation risk.

steilerDev:
**[product-architect]** APPROVED

Self-review for the record. Three-file infra chore; verified against the committed diff (`git diff origin/beta..HEAD`), not memory.

**AC1 — single konva on fresh install: PASS.** `npm ls konva --all` resolves to one node: `konva@10.3.0 overridden`, with `react-konva@19.2.4`'s transitive `konva` `deduped` to the same version. The override correctly forces a single copy. Lockfile is clean after install (no `package-lock.json`/`package.json` diff), and only one `node_modules/konva` entry exists in the lockfile — so the override is already reflected in the committed lockfile, no forward-looking-only drift.

**Guardrail robustness & placement: PASS.** `check-single-dep-version.sh` writes the `npm ls --json` tree to a tempfile (avoids argv/env size limits), walks it with a node ESM script, dedupes versions into a Set, and exits 1 when >1 distinct version resolves. Verified: 0-version and 1-version cases both exit 0 with a clear OK line; the multi-version path sets `FAILED=1 -> exit 1` by inspection. `npm ls` non-zero exits are correctly swallowed with `|| true` (npm ls exits non-zero on benign extraneous/peer states). CI step is well placed — in `static-analysis`, immediately after `npm ci` and before `npm audit signatures`, so it fails fast on a duplicate.

**Dep-policy compliance: PASS.** Exact pin `"konva": "10.3.0"` (no caret range), consistent with the existing exact pins in the `overrides` block.

**Architecture: no concerns.** The override is the right mechanism for collapsing a transitive duplicate, and the parameterized guardrail script is reusable for any future single-version invariant (just pass another package name). Proportionate, contained, no schema/API/contract surface touched.

Issue #1688, opened 2026-06-15
Diary overview: filters and search unusable on mobile (Filters toggle reveals off-screen panel)


steilerDev:
## Bug: Diary filters and search unusable on mobile — "Filters" toggle reveals an off-screen panel

**Reported by**: User
**Area**: Construction Diary (Bautagebuch) — Diary overview / `DiaryFilterBar`
**Severity**: High — core filtering and search are completely inaccessible on mobile viewports

### Problem

On mobile (`max-width: 767px`), tapping the "🔍 Filters" toggle in the diary overview appears to do nothing. The user cannot change the source-mode filter (all / manual / automatic) and cannot use the search input, because the panel that holds those controls is rendered off-screen.

### Expected behavior

On mobile, tapping the "🔍 Filters" button reveals the filter panel directly beneath the button, anchored to the filter bar. All controls — the mode chips (all / manual / automatic), the type chips, the date range, and the search input — are visible, on-screen, and operable. Tapping the button again hides the panel.

### Actual behavior

Tapping "🔍 Filters" toggles the `.filtersOpen` class (the panel switches from `display: none` to `display: grid`), but the panel renders off-screen and is unreachable. The user perceives the button as broken / non-functional, and neither the mode filters nor the search can be reached.

### Root cause (investigated by orchestrator)

- Component: `client/src/components/diary/DiaryFilterBar/DiaryFilterBar.tsx` + `client/src/components/diary/DiaryFilterBar/DiaryFilterBar.module.css`
- In the mobile breakpoint (`@media (max-width: 767px)`), `.filters` is positioned with:
  ```css
  .filters {
    position: absolute;
    top: 100%;
    left: 0;
    right: 0;
    ...
  }
  ```
- This panel is intended to anchor to its parent `.filterBar`, but `.filterBar` does **not** declare `position: relative`. (The `position: relative` at line 25 of the CSS module belongs to `.mobileToggle`, not `.filterBar`.)
- With no positioned `.filterBar` ancestor, the absolutely-positioned panel resolves its containing block to the nearest other positioned ancestor — an off-screen, visually-hidden `.liveRegion` (or the viewport) — so `top: 100%; left: 0; right: 0` place it far from the toggle button and effectively off-screen.
- Net effect: the panel is technically displayed but visually unreachable, so the mode chips and search input cannot be interacted with.

### Reproduction steps

1. Open the application on a mobile viewport (≤ 767px wide), or use browser dev tools device emulation.
[...]

PR #1689, opened 2026-06-15
fix(diary): anchor mobile filter panel to bar so filters and search are usable

steilerDev:
## Summary
- Fix CSS positioning of the diary filter panel on mobile so it anchors directly below the filter bar instead of overlapping the page header/content
- Add unit tests for `DiaryFilterBar` covering mobile panel visibility and positioning behavior
- Add new E2E spec (`diary-mobile-filters.spec.ts`) covering the mobile filter panel open/close flow and search interaction across all three viewports

Fixes #1688

## Test plan
- [ ] Unit tests pass (95%+ coverage on DiaryFilterBar)
- [ ] E2E tests pass on mobile, tablet, and desktop viewports


steilerDev (commented):
**[ux-designer]**

Reviewed CSS-only fix for Bug #1688 against the design system.

**Checklist**

- **Token adherence** — both new declarations (`position: relative` and `overflow: visible`) are non-visual layout properties; no color, spacing, radius, or font-size values were added. No token violations.
- **Desktop regression** — `position: relative` is placed on the top-level `.filterBar` rule, which applies at all breakpoints. This is safe: `position: relative` has no visual effect on desktop because the `.filters` panel is in normal flow there (not absolutely positioned). No layout regression.
- **Mobile dropdown anchor** — `position: absolute; top: 100%` on `.filters` now resolves against `.filterBar`, not an ancestor. The `overflow: visible` override in the `@media (max-width: 767px)` block ensures the dropdown is not clipped. Correct.
- **Dark mode** — only layout properties changed. All color tokens (`--color-bg-primary`, `--color-border`, `--shadow-md`) on `.filters` were pre-existing and already switch correctly in `[data-theme="dark"]`. Unaffected.
- **Accessibility** — `aria-expanded`, `role="group"`, `aria-pressed`, and all other ARIA attributes are in the TSX (not touched by this PR). The unit test additions cover `aria-expanded` transitions. No regression.
- **z-index** — the existing `z-index: var(--z-dropdown)` on `.filters` was not modified. Correct token still in use.

**Verdict: Approved.** The fix is minimal, correct, and introduces no design-system violations.

steilerDev (commented):
**[product-architect]** Verdict: APPROVE (posted as comment — cannot self-approve via API).

**Scope:** CSS-only bug fix (#1688) for the mobile diary filter panel rendering off-screen, plus unit + E2E coverage. No schema, API contract, or shared-type impact — no wiki updates required.

**Verified:**
- **Root cause fix is correct.** `.filters` is `position: absolute; top: 100%; left: 0; right: 0` on mobile. The absent positioning context on `.filterBar` made it anchor to a wrong ancestor / the initial containing block, so the panel rendered off-screen. Adding `position: relative` (line 10) establishes the correct containing block so the panel drops directly beneath the toggle. `overflow: visible` on the mobile breakpoint prevents the overflowing panel from being clipped. Minimal and targeted.
- **Token compliance.** Only layout properties (`position`, `overflow`) introduced — no hardcoded colors/spacing/radii. No design-token content changes, so no UX-designer gate triggered.
- **Unit coverage (DiaryFilterBar.test.tsx).** Six new scenarios exercise toggle presence, default-closed state, open/close on click, and `aria-expanded` transitions using accessible queries. 97.72% component coverage clears the 95% bar.
- **E2E coverage (diary-mobile-filters.spec.ts).** Six `@responsive` scenarios cover in-viewport panel visibility (bounding-box assertion), mode-chip reachability, search input, close-on-second-tap, aria-expanded, and the desktop/tablet inline-layout inverse. POM members (`mobileFilterToggle`, `searchInput`, `heading`, `DIARY_ROUTE`) all exist; `MOBILE_MAX_WIDTH = 767` matches the CSS breakpoint.

**Non-blocking note (informational):** the spec gates per-test via runtime `test.skip()` on viewport width rather than project-level filtering. Functionally correct and consistent with the existing diary E2E pattern — fine as-is.

No findings requiring changes.

Issue #1701, opened 2026-06-15
Disable LLM "thinking"/reasoning for budget extraction (fixes Gemini 2.5 timeout → LLM_UNREACHABLE)


steilerDev:
## Bug: Gemini 2.5 invoice extraction times out (LLM_UNREACHABLE) because model "thinking" is enabled by default

**Type**: Bug
**Priority**: Should Have
**Component**: Budget invoice auto-itemize (LLM budget extraction)
**Parent Epic**: None — auto-itemize was delivered as standalone stories (#1545–#1547); related epics (EPIC-15, EPIC-05) are closed.

### Problem

When the budget invoice auto-itemize feature is configured with **Gemini 2.5 Flash** via the OpenAI-compatible endpoint `https://generativelanguage.googleapis.com/v1beta/openai/`, invoice extraction requests time out after `LLM_REQUEST_TIMEOUT_MS` (default 30000ms) and surface to the user as:

```json
{ "error": { "code": "LLM_UNREACHABLE", "message": "LLM provider is unreachable" } }
```

The server log shows the underlying cause is a clean `AbortError` with **no network `innerCause`** — i.e., the request connects fine and the provider is reachable, but the model does not respond within the timeout. The `LLM_UNREACHABLE` code is misleading: the provider is reachable; the model is simply too slow.

### Root cause

Gemini 2.5 models (including Flash) have **"thinking"/reasoning enabled by default** with a dynamic thinking budget. The extraction call is **non-streaming** and `LLM_MAX_TOKENS` defaults to **16384**, so the model spends significant time on chain-of-thought reasoning before emitting the structured JSON, pushing total latency past 30s on real-world invoices.

The request body is assembled in `buildRequestBody()` in `server/src/services/budgetExtraction/providerProfiles.ts`. It currently sends only `model`, `messages`, `temperature: 0`, `max_tokens`, and (per provider) `response_format` — **no parameter to disable model thinking**. The timeout/abort is then mapped to `LlmUnreachableError` / `LLM_UNREACHABLE` in `server/src/services/budgetExtraction/openAICompatibleProvider.ts` (the `catch` block around the `fetch` call).

### Expected behavior

- Invoice line-item extraction is a **structured-extraction task** that does not require chain-of-thought reasoning. Model "thinking"/reasoning should be **disabled** for the budget-extraction call, reducing latency (fixing the timeout at its source) and lowering token cost.
[...]

PR #1702, opened 2026-06-15
fix(llm): disable model thinking/reasoning for budget invoice extraction

steilerDev:
## Summary

- Gemini 2.5 Flash has dynamic thinking enabled by default on its OpenAI-compatible endpoint, causing 30s+ timeouts that surfaced as `LLM_UNREACHABLE` errors during budget invoice extraction
- Added per-provider reasoning suppression in `buildRequestBody()`: Gemini and Ollama unconditionally get `reasoning_effort: "none"`; OpenAI only receives it for recognized reasoning models (o1/o3/o4/gpt-5/codex-mini series) to avoid HTTP 400 on non-reasoning models like gpt-4o; Anthropic and generic providers are unchanged
- Added `isOpenAiReasoningModel()` helper with conservative pattern matching (prefers false negatives over false positives to avoid HTTP 400)

Fixes #1701

## Test plan

- [x] Unit tests for `isOpenAiReasoningModel()`: reasoning models (o1, o3-mini, o4-turbo, gpt-5, codex-mini) return `true`; non-reasoning models (gpt-4o, gpt-3.5-turbo) return `false`; unknown variants return `false`
- [x] Unit tests for `buildRequestBody()`: verify `reasoning_effort: "none"` is present for gemini/ollama providers, only for OpenAI reasoning model IDs, and absent for anthropic/generic
- [x] Pre-commit hook quality gates pass

### DIFF
--- a/server/src/services/budgetExtraction/providerProfiles.test.ts
+++ b/server/src/services/budgetExtraction/providerProfiles.test.ts
@@ -8,6 +8,7 @@ import {
   detectProvider,
   parseProviderEnv,
   buildRequestBody,
+  isOpenAiReasoningModel,
   LLM_PROVIDERS,
 } from './providerProfiles.js';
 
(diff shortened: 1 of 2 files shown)


steilerDev:
**[product-architect]** Architecture & code-quality review of the thinking-disable change in `buildRequestBody()`.

## Verdict: APPROVE (with non-blocking findings to consider before merge)

The change is well-scoped, sits at the correct layer, and is properly tested. It maps cleanly onto the ACs (AC1 gemini, AC2 openai-conditional, AC3 anthropic-unchanged). No blocking architectural objections.

### 1. Per-provider approach & name-based OpenAI detection — sound for this scale

Name-based reasoning-model detection is an **acceptable architectural choice here**, not a config-smell. The alternative (a new env var to declare "is reasoning model") pushes provider-internals onto the operator for a single-instance, <5-user app and contradicts our "simplicity first" principle. The provider already self-describes via `model`; deriving capability from it is reasonable. Keeping the detection in a small, exported, pure, unit-tested helper is the right containment. Good call exporting it for direct testing.

### 2. Right layer — yes

`buildRequestBody()` is exactly where provider-divergent request shaping already lives (`response_format` per provider). Co-locating the thinking-disable field with the existing per-provider switch keeps all wire-shape divergence in one function. No reason to push this into the provider/transport layer.

### 3. Non-blocking findings

**(a) [Medium] `reasoning_effort: 'none'` value validity on o1/o3/o4 may not match the field's accepted enum.** The detection correctly gates *whether the field is sent*, but the *value* `'none'` is a gpt-5-era addition. Classic o-series (o1/o3/o4) historically accept `low|medium|high` and may reject `'none'` with a 400 — meaning a `gpt-4o` user is safe (field omitted) but an `o3-mini` user could now get a 400 where before they got a slow-but-working call. The fix's own rationale ("false positives would cause a 400") applies to the *value* too, not just field presence. Since the concrete bug being fixed is Gemini, consider whether OpenAI reasoning models even need `'none'` here, or whether `'low'` is the safer minimal value for the o-series. Worth a quick confirmation against current OpenAI docs before relying on it.

**(b) [Low] Ollama now receives `reasoning_effort: 'none'` — mild deviation from AC4's "unchanged."** AC4 asks ollama/generic to stay unchanged "unless a safe, provider-detected mechanism exists." The code comment asserts it's "safely ignored on non-thinking models," but OpenAI-compat servers vary in strictness about unknown fields (some 400 on unrecognized params). This is the riskiest of the unconditional additions because "ollama" covers arbitrary self-hosted gateways. If we want to honor AC4 literally, the conservative move is to omit it for ollama (like generic). Not blocking — but document the assumption or narrow it.

**(c) [Low / maintainability] Regex will need extension for future families (o5, etc.).** Acknowledged in the issue. `/^o[134](-|$)/` deliberately omits o2 (never existed) and o5 (future). This is fine — a false negative is safe by design (slow call, not a 400). The doc comment already calls this out. No change needed, just confirming the tradeoff is intentional and the failure mode is benign.

### 4. Test coverage — adequate

57/57 pass, 100% file coverage. `isOpenAiReasoningModel` has good positive/negative tables including case-insensitivity and the `gpt-4o` false-positive guard. The `buildRequestBody` tests assert both presence/absence of `reasoning_effort` and preservation of `response_format`, satisfying AC5. One coverage gap matching finding (a): there's no test pinning the *value* sent to o-series specifically vs gpt-5 — but since the code sends the same `'none'` for all, that's a code question, not a test gap.

### Summary
Approve. Findings (a) and (b) are runtime-behavior risks worth a 2-minute doc check before merge, but neither breaks a contract nor the documented bug fix (Gemini), so they don't block.

Issue #1707, opened 2026-06-16
Mobile: prevent viewport auto-zoom on input focus and disable pinch-to-zoom


steilerDev:
**[product-owner]**

## Problem

On mobile (notably iOS Safari), the viewport auto-adjusts/zooms when a user taps into a text field, and pinch-to-zoom is possible anywhere in the app. Both behaviors break the native-app illusion Cornerstone aims for as a self-hosted PWA-style app for 1-5 homeowners.

The current viewport meta tag in `client/index.html` is:

```html
<meta name="viewport" content="width=device-width, initial-scale=1.0" />
```

This permits both browser auto-zoom on input focus and user-initiated pinch-to-zoom.

User report (verbatim):
> On mobile, when highlighting a field, often the viewport adjusts to focus on the text field, breaking the app illusion. It should not be possible to pinch to zoom on the web UI — is there a way to prevent this?

## Expected behavior

- Focusing an input/textarea/select on mobile does **not** trigger the browser to zoom the viewport.
- Pinch-to-zoom is disabled across the entire app on all routes/pages.
- The app behaves like a fixed-scale native application on mobile.

## Actual behavior

- Tapping into a text field on iOS Safari auto-zooms the viewport to focus on the field.
- Users can pinch to zoom in/out anywhere in the UI.

## Reproduction steps

**Auto-zoom on focus (iOS Safari):**
1. Open the app on an iPhone in Safari.
2. Navigate to any page with a form (e.g., create/edit work item, login).
3. Tap into a text input or textarea.
4. Observe: the viewport zooms in to center the focused field.

**Pinch-to-zoom:**
1. Open the app on any mobile device.
2. On any page, place two fingers on the screen and pinch outward or inward.
3. Observe: the page zooms, scaling the entire UI.

## Acceptance Criteria (Given/When/Then)

1. **Given** a user is viewing any page on a mobile device, **When** they perform a pinch gesture (two fingers, in or out), **Then** the page does not zoom and remains at the fixed scale.

2. **Given** a user is on iOS Safari, **When** they tap into any text input, textarea, or select element, **Then** the browser does not auto-zoom the viewport.

3. **Given** the app is loaded on any device, **When** the viewport meta tag in `client/index.html` is inspected, **Then** it disables user scaling and caps the maximum scale (e.g., `maximum-scale=1, user-scalable=no` in addition to `width=device-width, initial-scale=1`).

[...]

PR #1709, opened 2026-06-16
fix(client): disable mobile viewport zoom on input focus and pinch (#1707)

steilerDev:
## Summary

- Added `maximum-scale=1, user-scalable=no` to the viewport meta tag in `client/index.html`
- Prevents iOS Safari from auto-zooming when a form input is focused
- Disables pinch/double-tap zoom to preserve the native-app feel on mobile

Fixes #1707

## Test plan

- [ ] Unit tests pass (95%+ coverage)
- [ ] Integration tests pass

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1, user-scalable=no" />
     <meta name="description" content="Cornerstone - Home Building Project Management" />
     <title>Cornerstone</title>
     <link rel="icon" type="image/svg+xml" href="/favicon.svg" />


steilerDev:
**[ux-designer]**

## Design Review — PR #1709 (Viewport zoom fix, Issue #1707)

### Accessibility Tradeoff Verdict: RECOMMEND NARROWER APPROACH

`user-scalable=no` / `maximum-scale=1` is a WCAG 1.4.4 violation (Resize Text, Level AA). While the WCAG specification does carve out a known exception for "user agents", browser vendors have deliberately closed that loophole — iOS Safari 10+, Chrome 32+, and Firefox all ignore `user-scalable=no` in the viewport meta tag when the user has enabled accessibility zoom in system settings. This means the directive does not reliably prevent pinch-to-zoom for users who need it, while still failing a WCAG audit.

More importantly, **the root cause of the iOS auto-zoom is solvable without disabling pinch-to-zoom at all.** iOS Safari auto-zooms on `<input>` focus when the input's computed `font-size` is below 16px. The current `.input` class in `shared.module.css` uses `font-size: var(--font-size-sm)` which resolves to **14px** — below the threshold. Bumping inputs to 16px on mobile eliminates the auto-zoom trigger entirely.

### Recommendation

Keep `initial-scale=1.0` and `width=device-width`. Drop `maximum-scale=1, user-scalable=no` entirely. Instead, add a mobile-scoped CSS rule to `shared.module.css` (or `index.css`) that overrides input font-size to `--font-size-base` (16px) on mobile:

```css
@media (max-width: 639px) {
  .input,
  select.input,
  textarea.input {
    font-size: var(--font-size-base); /* 16px — prevents iOS Safari auto-zoom on focus */
  }
}
```

This:
- Eliminates the iOS auto-zoom without disabling pinch-to-zoom
- Preserves WCAG 1.4.4 compliance
- Works in the same mobile contexts where auto-zoom is a problem

The only behavioural difference is that input text is rendered at 16px instead of 14px on screens narrower than 640px — an acceptable and visually appropriate change for mobile.

If the product decision is that pinch-to-zoom must be disabled regardless (native-app feel), the current change is the correct implementation. In that case this finding is **non-blocking** given the acknowledged tradeoff for a 1–5 user self-hosted app where users can configure their OS zoom preference. But the narrower CSS fix is worth doing first — it solves the actual reported bug without the WCAG trade.

**Verdict: Non-blocking. Recommend applying the CSS input font-size fix instead and reverting the viewport directive change. If product explicitly accepts the a11y tradeoff, approve as-is.**

Issue #1705, opened 2026-06-16
Photo annotation is broken on mobile/touch devices (no fit-to-container, no touch input)


steilerDev:
**As a** homeowner using Cornerstone on a phone or tablet, **I want** to annotate construction photos with the same tools available on desktop **so that** I can mark up site photos directly from the device I take them on.

**Parent Epic**: #1472 (Photo annotation editor — Shottr-style)
**Priority**: Should Have
**Type**: Bug

---

### Problem

The photo annotation editor is effectively unusable on touch devices. Users cannot reliably see the photo they are annotating, and the annotation tools (pen, shapes, text, measurement) do not respond to touch input at all. This breaks a core promised capability — marking up photos — for everyone working from a phone or tablet, which is the primary device for capturing construction-site photos.

To directly answer the reported question ("Is special support required to enable this on mobile?"): **yes.** Touch/mobile is not a free side effect of the desktop build. The canvas must be explicitly scaled to fit its on-screen container, and touch/pointer input must be explicitly wired to the drawing tools. Neither is currently implemented.

### Expected behavior

- On any device (phone, tablet, desktop), the photo renders **fully fit to its container** — the whole image is visible, scaled down to fit the available space, never clipped and never zoomed past 1:1.
- All annotation tools (pen/freehand, shapes, arrows, text, measurement) are fully operable **by touch** on phone and tablet, matching the desktop tool behavior.
- No excessive/unintended zoom on initial open at any viewport.
- Desktop mouse behavior is unchanged.

### Actual behavior

- **On a phone:** the photo does not render / is effectively invisible. The annotation canvas is sized to the photo's intrinsic pixel dimensions, which is far larger than the small phone viewport, so the visible image is clipped out of view.
- **On a tablet:** the photo opens dramatically over-zoomed (showing only a small region of the image), because the canvas is again sized to the photo's full pixel dimensions rather than fit to the container.
- **On any touch device (phone and tablet):** none of the drawing tools do anything when used with a finger/stylus — tapping or dragging on the canvas produces no annotation.
- **On desktop:** annotation works as expected (this regression is touch/responsive-only).

### Reproduction steps

1. On a phone, open a photo (e.g., from the construction diary or a work item) and tap to annotate it.
2. Observe: the photo does not appear / is clipped out of the visible area.
3. Switch to a tablet, open the same annotation editor.
4. Observe: the photo appears massively zoomed-in, showing only a fraction of the image.
[...]

PR #1711, opened 2026-06-16
fix(photo-annotator): responsive scaling and touch support for mobile annotation

steilerDev:
## Summary

Fixes the photo annotation editor to be responsive on mobile/tablet devices and enables touch drawing by implementing responsive scaling with ResizeObserver.

### `client/src/lib/konvaInit.ts`
- Added `Konva.hitOnDragEnabled = true` for better touch event handling

### `client/src/components/photos/PhotoAnnotator/PhotoAnnotator.module.css`
- Added `touch-action: none` to `.canvasArea` to prevent browser touch defaults

### `client/src/components/photos/PhotoAnnotator/PhotoAnnotator.tsx`
- Added `canvasAreaRef` and `containerSize` state to track container dimensions
- Added ResizeObserver effect to measure container size and compute fitScale
  - fitScale = min(containerWidth / intrinsicWidth, containerHeight / intrinsicHeight, 1.0)
  - 1.0 cap prevents excessive zoom on small photos
- Added pointer capture effect to enable reliable multi-touch drawing
- Updated Stage dimensions:
  - `width/height` = intrinsic dimensions × fitScale (for CSS sizing)
  - `scaleX/scaleY` = fitScale (Konva's scaling)
  - This makes `getPointerPosition()` return intrinsic-space coordinates
- Updated KonvaImage to use intrinsic dimensions (not scaled)
- Renamed all Stage handlers from mouse to pointer events:
  - `onMouseDown` → `onPointerDown`
  - `onMouseMove` → `onPointerMove`
  - `onMouseUp` → `onPointerUp`
  - Changed event type from `MouseEvent` to `PointerEvent`
- Added `ref={canvasAreaRef}` to canvas container div

## Technical Details

All shape coordinates remain in intrinsic photo-pixel space with no serialisation format changes. The scaling is applied only at the Konva Stage level. The inline text input already correctly computes `scale = stageRect.width / (photo.width ?? 800)`, which now equals fitScale and requires no changes.

Both new effects include proper cleanup:
- ResizeObserver: calls `ro.disconnect()`
- Pointer capture: calls `removeEventListener`

Fixes #1705


steilerDev:
**[ux-designer]**

Design/UX review for #1705 (mobile annotation fix). Reviewing: `PhotoAnnotator.tsx`, `PhotoAnnotator.module.css`, `ToolPalette.module.css`, `konvaInit.ts`.

---

## Verdict: APPROVE

This is a well-scoped bug fix. All five UX checkpoints pass.

---

### 1. Fit-to-container scaling — PASS

The approach is correct:

- `fitScale = min(containerWidth / intrinsicW, containerHeight / intrinsicH, 1.0)` ensures the photo fills the available space without exceeding native resolution.
- The Stage receives `width={intrinsicW * fitScale}` / `height={intrinsicH * fitScale}` and `scaleX/scaleY={fitScale}`, while `KonvaImage` keeps `width={intrinsicW}` / `height={intrinsicH}`. This is the standard Konva fit-to-container pattern: the Stage viewport matches the container, the image inside it is at native resolution, and the scale transform bridges the two. No clipping on phone, no upscaling on tablet.
- The `1.0` cap on `fitScale` prevents small photos from being stretched to fill large containers — correct, pixel-for-pixel fidelity is preserved for small images.
- ResizeObserver with `deps=[imageLoaded]` correctly defers attachment until the canvas div is mounted (the `if (!el) return` guard handles the loading state).

### 2. ToolPalette touch targets — PASS

`.toolButton` has `min-width: 44px; min-height: 44px` (ToolPalette.module.css lines 34–35). `.swatchButton` uses `width: 24px; height: 24px` with `padding: 10px; box-sizing: content-box`, giving a 44×44px touch target (24 + 20 = 44). Both meet the 44×44px minimum. No regressions from this PR.

### 3. `touch-action: none` scoping — PASS

The property is applied only to `.canvasArea` (line 16) and `.svgOverlay` (line 31) — both are within the annotator canvas area, not on a scroll container or page shell. Browser scrolling is unaffected outside the annotator. The existing `touch-action: none` on `.svgOverlay` was already in place; the new one on `.canvasArea` is the correct addition.

### 4. Accessibility with scaled canvas — PASS

`role="application"` with `aria-label={t('canvas')}` is unchanged. Scaling is a visual transform on the Stage; the `role="application"` wrapper div retains its dimensions (it is a `flex: 1` container, not the Stage element itself), so the ARIA landmark is unaffected. Keyboard/screen-reader users interact with the palette buttons, not the canvas directly — no regression.

### 5. Token adherence — PASS

The only CSS change in this PR is the single line `touch-action: none` added to `.canvasArea`. No hardcoded colors, spacing, radii, or font sizes were introduced.

---

### Informational (pre-existing, not introduced by this PR)

- `ToolPalette.module.css` line 112: `.sizeDropdownSelect:focus-visible` uses `outline: 2px solid var(--color-focus-ring)` instead of the project convention `box-shadow: var(--shadow-focus)`. This inconsistency predates this PR — worth a refinement pass alongside the next annotator story.
- `PhotoAnnotator.module.css` line 43: `.inlineInput { z-index: 1000 }` should be `var(--z-modal)`. Also pre-existing; tracked in agent memory.

Both are refinement items, not blockers for this fix.

steilerDev:
**[product-architect]** Architecture & code-quality review (bug fix #1705)

**VERDICT: CHANGES REQUESTED** — one correctness defect in the core fit-to-container path.

## Critical: pointer coordinates are not mapped back into intrinsic space

The Stage is scaled (`scaleX/scaleY={fitScale}`) while `KonvaImage` and all shapes stay in intrinsic coordinate space — that part of the design is sound. But the new draft-drawing path stores raw `stage.getPointerPosition()` values directly as intrinsic shape coordinates:

- `PhotoAnnotator.tsx` L431/477–481 (`startX/startY/endX/endY`, freehand first point)
- L497/502/505 (move: freehand point push, rect/line `endX/endY`)

`stage.getPointerPosition()` does **not** invert the Stage's `scale` attribute. Per Konva `Stage._getContentPosition()` (node_modules/konva/lib/Stage.js:649–665), it only divides by the container's *CSS* scale (`rect.width / content.clientWidth`), which is `1` here because the container div is sized to `stageWidth = intrinsicW * fitScale` with no CSS transform. So `getPointerPosition()` returns values in `0…intrinsicW*fitScale` (screen/container space), but they are stored as intrinsic coordinates in `0…intrinsicW`.

Effect: whenever `fitScale < 1` — i.e. the entire mobile/large-photo case this PR targets — every newly drawn shape lands compressed into the top-left `fitScale` fraction of the image, far from where the user touched/clicked. A touch at the true bottom-right corner is stored at ~`fitScale` of the way across.

**Fix:** map the pointer back through the inverse Stage transform before storing. Either:
- use `stage.getRelativePointerPosition()` (Konva `Node.getRelativePointerPosition`, node_modules/konva/lib/Node.js:629) instead of `getPointerPosition()`, or
- divide the returned `pos.x/pos.y` by `fitScale`.

Note the existing endpoint-drag handlers (L847–884) are already correct — they read `e.target.position()`, which is node-local (intrinsic) space, so only the `getPointerPosition()` draft path needs the change.

## Test coverage gap (related to the above)

The new unit tests verify Stage `width/height/scaleX/scaleY` attributes (good) and exercise the draw handlers, but the react-konva mock's `getPointerPosition()` returns a fixed value with no scale modelling, so the round-trip "screen pointer → stored intrinsic coordinate" is never asserted under `fitScale < 1`. The drawing test (L1204+) sets pointer `(50,50)` and expects the draft to use `(50,50)` verbatim — which would still pass with the bug present. Please add a case at `fitScale = 0.5` asserting the stored coordinate is the de-scaled value (after the fix, the mock should model the inverse transform, or the test should assert the divided value).

The ~34% file coverage baseline for Konva render helpers is fine and not a blocker — that is pre-existing and not JSDOM-testable.

## Looks good

- Fit-to-container approach (measure `.canvasArea` via ResizeObserver, `fitScale = min(cW/iW, cH/iH, 1.0)`, Stage size = intrinsic×fitScale, KonvaImage/shapes in intrinsic space, serialisation unchanged) — sound and minimally invasive.
- ResizeObserver effect cleanup (`ro.disconnect()`) is correct; re-running on `imageLoaded` is appropriate.
- Inline text-input overlay math (L710–767) is correct and robust — it derives `scale` from the live DOM `getBoundingClientRect()` rather than from `getPointerPosition()`, so it is unaffected by the bug above.
- Mouse+touch handlers wired to shared callbacks, `preventDefault()` on touch, `touch-action: none`, and `Konva.hitOnDragEnabled = true` — all appropriate for mobile.
- TS conventions clean: no unjustified `any` in production code (the `any` in the mock is justified + eslint-disabled), `MouseEvent | TouchEvent` typing is correct, ESM/type-import conventions followed.

No schema or API-contract surface touched — no wiki updates required for this change.

steilerDev:
**[product-architect]** Verdict updated: **APPROVE**.

My earlier blocking finding (drawing handlers used `stage.getPointerPosition()`, which does not invert the Stage's `scaleX/scaleY=fitScale`, causing shapes to land compressed in the top-left when `fitScale < 1`) is fully resolved on `939d2af2`.

Verified:
- **Production fix** — both drawing-handler call sites use `getRelativePointerPosition()` (`handleStagePointerDown` L431, `handleStagePointerMove` L497). Repo-wide grep confirms no stray `getPointerPosition()` remains for shape coordinates; remaining references are test comments only. `getRelativePointerPosition()` correctly returns intrinsic image-space coords because it inverts the Stage transform (scale + position), which is exactly what shape persistence requires.
- **Mock** — `__mocks__/react-konva.ts` now exposes `getPointerPosition()` and `getRelativePointerPosition()` as independently-settable values returning *different* coordinates, enabling the test to discriminate which method production calls.
- **Regression guard is live, not a no-op** — confirmed the committed rectangle is actually surfaced in `data-annotator-shapes` (intrinsic `x:100,y:100,w:200,h:100`), so the discriminating assertions execute. A mutation check (temporarily reverting production to `getPointerPosition()`) makes test 6 fail with `Received: 50` vs `Expected: 100` under `fitScale=0.5` — proving the test catches the precise scale-inversion regression. The `.not.toBeCloseTo(50)` negative assertions bite as intended.

No remaining architectural concerns. Approving.

Issue #1708, opened 2026-06-16
fix(search-picker): dropdown detaches from input field on mobile scroll


steilerDev:
**Type:** Bug (confirmed, mobile UI)

## Summary

On mobile, the `SearchPicker` dropdown (used by the Area picker, Orientation picker, and other entity selectors) detaches from its originating text field when the user scrolls the page while the dropdown is open. The dropdown should stay anchored to (fixed relative to) its originating text field during scroll.

## Technical Root Cause

(From code investigation — `client/src/components/SearchPicker/SearchPicker.tsx`)

- The dropdown menu is rendered via `ReactDOM.createPortal` into `document.body` with `position: fixed`, using coordinates from `inputRef.current.getBoundingClientRect()`.
- The portal/`fixed` approach was introduced in **PR #1601** specifically to escape modal overflow clipping. **The fix must NOT regress that** — keep the portal; do not move the menu back inside an overflow-clipped container.
- Position is recalculated on a capture-phase `window` `scroll` listener (`updateDropdownRect`, lines ~114-127). On desktop this tracks the field. On mobile touch/momentum scrolling, those discrete scroll events fire unreliably and do not account for the **visual viewport**, so the `position: fixed` dropdown keeps stale coordinates and visually floats away from the field.

## Expected Behavior

While the dropdown is open and the user scrolls (on mobile **and** desktop), the dropdown remains visually anchored to its originating text field.

## Acceptance Criteria

- [ ] **Given** a SearchPicker (Area or Orientation picker) is open on a mobile viewport, **When** the user scrolls the page, **Then** the dropdown menu stays anchored to the input field (moves together with the field) without lag/detachment.
- [ ] **Given** the SearchPicker is inside a modal/dialog with overflow, **When** the dropdown opens, **Then** it is not clipped by the modal (the #1601 fix is preserved).
- [ ] **Given** the dropdown is open and the field scrolls fully out of view, **Then** the dropdown behaves sensibly (stays anchored / scrolls away with the field, not stuck mid-viewport).

## Notes

- Parent epic EPIC-16 (#752, Floor Plans & Utility Tracking), which introduced the Area/Orientation pickers, is closed — filing as a standalone bug.
- Affects all SearchPicker consumers, not just the floor-plan pickers.

steilerDev:
[...]

PR #1712, opened 2026-06-16
fix(search-picker): anchor dropdown to input field during mobile scroll

steilerDev:
## Summary

- Fixes the mobile bug where the `SearchPicker` dropdown (Area picker, Orientation picker, and all other consumers) detached from its originating input field when the user scrolled the page with the dropdown open.
- The dropdown now tracks its input field every paint frame via a `requestAnimationFrame` loop while open, so it stays anchored during mobile touch/momentum scrolling (where coalesced `scroll` events left the `position: fixed` portal with stale coordinates). A `closeIfOutOfView` guard closes the dropdown when the field scrolls fully out of the viewport.
- An equality guard in `updateDropdownRect` returns the previous `DOMRect` reference when the field hasn't moved, avoiding ~60fps re-renders while the dropdown is idle-open.
- The portal-into-`document.body` + `position: fixed` approach from #1601 (anti-clipping inside modals) is preserved.

Fixes #1708

## Test plan
- [x] Unit tests pass (57 tests; SearchPicker.tsx coverage 96.9% statements / 100% functions). New `#1708` tests cover the rAF loop lifecycle, the close-on-out-of-view guard (above/below viewport), and re-render suppression when the rect is unchanged.
- [x] New mobile-viewport E2E smoke test asserts the dropdown renders anchored near its input on a mobile viewport.
- [ ] CI Quality Gates pass

### DIFF
--- a/.claude/agent-memory/e2e-test-engineer/MEMORY.md
+++ b/.claude/agent-memory/e2e-test-engineer/MEMORY.md
@@ -1,7 +1,15 @@
 # E2E Test Engineer — Agent Memory (Index)
 
 > Detailed notes live in topic files. This index links to them.
-> See: `e2e-pom-patterns.md`, `e2e-parallel-isolation.md`, `story-epic08-e2e.md`, `story-933-dav-vendor-contacts.md`, `milestones-e2e.md`, `story-1248-mass-move.md`, `photo-annotator-e2e.md`
+> See: `e2e-pom-patterns.md`, `e2e-parallel-isolation.md`, `story-epic08-e2e.md`, `story-933-dav-vendor-contacts.md`, `milestones-e2e.md`, `story-1248-mass-move.md`, `photo-annotator-e2e.md`, `searchpicker-mobile-1708.md`
+
+## SearchPicker mobile anchor regression (Issue #1708, 2026-06-16) — `e2e/tests/responsive/search-picker-mobile.spec.ts`
+
+- Scenario 1 (@responsive, mobile-only MOBILE_MAX_WIDTH=499): focuses AreaPicker on WorkItemCreate, asserts `[data-search-picker-dropdown]` visible, `Math.abs(dropdownBox.y - inputBottom) < 20`.
+- Scenario 2: `test.skip(true)` — no lightweight modal+picker E2E fixture; covered by SearchPicker.test.tsx unit test.
+- AreaPicker uses `showItemsOnFocus={true}` → dropdown opens on `.click()` with no typing needed.
+- Dropdown closes after `.click()` on an option — asserted via `not.toBeVisible()`.
+- See `searchpicker-mobile-1708.md` for full context.
 
 ## OrientationsTab E2E coverage (fix #1687, 2026-06-15) — `e2e/tests/orientations.spec.ts`, `e2e/pages/OrientationsPage.ts`
 
--- a/.claude/agent-memory/e2e-test-engineer/searchpicker-mobile-1708.md
+++ b/.claude/agent-memory/e2e-test-engineer/searchpicker-mobile-1708.md
@@ -1,0 +1,28 @@
+---
+name: searchpicker-mobile-1708
+description: SearchPicker mobile anchor regression test details for Issue #1708 — rAF position-tracking fix
+metadata:
+  type: project
+---
+
+# SearchPicker Mobile Anchor Regression (Issue #1708)
+
+**Fix:** `SearchPicker.tsx` now runs a `requestAnimationFrame` loop while the dropdown is open to call `updateDropdownRect()` (syncs `position:fixed` dropdown with input's `getBoundingClientRect()`) and `closeIfOutOfView()` (closes if input scrolls off screen). Replaces the prior scroll/resize listeners which missed momentum-scroll frames on mobile WebKit.
+
+**Test surface used:** `WorkItemCreate` page (`/project/work-items/new`) — AreaPicker (`SearchPicker<TreeNode>`) with `showItemsOnFocus={true}`.
+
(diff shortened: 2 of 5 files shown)


steilerDev:
**[product-architect]**

## Architecture Review — PR #1712 (issue #1708)

**Verdict: APPROVE** (1 medium observation, noted for refinement — not blocking)

Reviewed for architecture compliance, code quality, and test coverage. This is a contained, frontend-only fix to `SearchPicker` — no schema, API contract, or shared-type impact. The portal + `position: fixed` anti-clipping behavior from #1601 is correctly preserved.

### What was verified

- **rAF approach is appropriate.** Mobile momentum/fling scrolling coalesces `scroll` events and does not fire one per animation frame, so a `requestAnimationFrame` loop is the standard, correct way to keep a `position: fixed` portal anchored to a scrolling input. The retained `scroll`(capture)/`resize` listeners give immediate response on desktop. Good layered design.
- **Effect lifecycle/cleanup is correct.** The loop is started in the `isOpen` effect and torn down in its cleanup (`cancelAnimationFrame` + listener removal, with `rafRef` nulled). A second unmount-only effect provides defense-in-depth for both the rAF handle and the debounce timer. No leak path identified — closing via `closeIfOutOfView` flips `isOpen`, which re-runs the effect cleanup.
- **Re-render guard is sound.** The functional-updater equality check in `updateDropdownRect` (top/left/width/bottom) returns the previous reference when unchanged, so a stationary open dropdown does not re-render every frame. Test E explicitly proves zero extra renders across two callback invocations. The `eslint-disable` for `set-state-in-effect` is justified and correctly scoped (functional updater, invoked from rAF not directly in an effect body).
- **Test coverage is thorough and proportional.** The new `#1708` describe block covers: loop start on open / cancel on close (A), both out-of-view branches — bottom<0 (B) and top>innerHeight (C), the in-view stay-open path (D), and the no-re-render guard (E). The rAF spy that captures rather than recursively invokes the callback is the right technique for deterministic assertions. E2E adds a mobile anchor-proximity smoke test and documents why true momentum-scroll and the modal+picker case are covered by unit tests instead — a reasonable, well-justified scoping decision.

### Medium observation (non-blocking, for refinement)

- **Per-frame forced reflow while open.** The loop calls `getBoundingClientRect()` twice per frame (once in `updateDropdownRect`, once in `closeIfOutOfView`) for the entire time the dropdown is open, even when nothing is scrolling. At this app's scale (<5 users, one open picker at a time) this is acceptable and not worth gating the fix. Two cheap future refinements if it ever matters: (1) dedupe to a single `getBoundingClientRect()` call per frame, sharing the rect between position-update and out-of-view logic; (2) consider gating the rAF loop behind an active-scroll signal (e.g. start on `scroll`/`touchmove`, idle-stop after N quiet frames) rather than running unconditionally. Neither is required now.

No contract, naming, or pattern violations. Cleanup, error handling, and typing are all correct. Approving.

steilerDev:
**[ux-designer]**

## Design Review — PR #1712 (Issue #1708): SearchPicker mobile scroll anchor

### Summary

This is a behavioral-only change. No CSS, design tokens, or markup structure were modified. The diff touches one production file (`SearchPicker.tsx`), a unit test file, and a new E2E spec. There are no token regressions, dark mode regressions, or styling changes to audit.

---

### UX Assessment: rAF loop for scroll tracking

**Close-on-out-of-view: Correct UX choice.**

Closing the dropdown when its anchor field scrolls off-screen is the right call. A floating `position: fixed` dropdown that remains visible after its input has scrolled away is disorienting — the user has no way to correlate the floating list with any field. iOS Safari and Android Chrome both use this pattern for native `<select>` popups. The ±0px/±100% viewport boundary threshold (`rect.bottom < 0 || rect.top > window.innerHeight`) is exactly the right trigger: it closes only when the input is fully out of view, not prematurely while it is still partially visible during a slow scroll.

**rAF-based position tracking: Correct approach for mobile.**

The `scroll` event is coalesced during momentum/fling scrolling on mobile WebKit — it fires at the start of inertia scrolling but not on every repaint frame. A `requestAnimationFrame` loop reading `getBoundingClientRect()` per frame is the standard remedy. The existing `scroll` listener is kept for non-mobile paths where per-frame polling would be wasteful, so the two mechanisms complement each other. The render-skipping optimization (functional updater that bails if top/left/width/bottom are unchanged) prevents the per-frame polling from causing a re-render on every animation frame when the user is not scrolling — this is important for battery life and scroll smoothness on mobile.

**One concern: continuous rAF loop on desktop.**

The rAF loop fires on every animation frame (roughly 60 fps) for the entire lifetime of an open dropdown, including on desktop where `getBoundingClientRect()` is cheap but the polling is unnecessary. The scroll/resize listeners on desktop are sufficient. A future refinement could gate the loop on a `matchMedia('(pointer: coarse)')` check or a detected touch event to limit polling to touch devices. This is a non-blocking suggestion — the battery/CPU impact is marginal given the dropdown is typically open for only a few seconds, and correct behavior on mobile outweighs the micro-optimization.

---

### Token / Styling Compliance

No CSS changes. No hardcoded values introduced. No token regressions. Pass.

### Dark Mode

No color or surface changes. Pass.

### Responsive Behavior

The fix specifically targets mobile viewports. The E2E spec correctly gates the anchor-proximity assertion to `viewportWidth <= 499px`. Pass.

### Accessibility

No ARIA, focus management, or keyboard navigation changes. The `Escape` key handler on the dropdown is unchanged. The `closeIfOutOfView` path calls `setIsOpen(false)` directly — this is correct because the input itself is off-screen, so focus has already moved or the user has deliberately scrolled away; there is no focus management to perform. Pass.

**One informational note:** when the dropdown closes due to scroll-out-of-view, no screen reader announcement is made (e.g. via a live region). This mirrors the behavior of native browser `<select>` elements and is acceptable — announcing every scroll-triggered dismissal would be noisy and unhelpful. No change required.

---

### Verdict

**Approved.** The fix is the correct UX and technical approach. No token, dark mode, accessibility, or responsive regressions. The continuous rAF-on-desktop observation is informational only and does not block merge.

Issue #1744, opened 2026-06-17
Attach documents to budget sources and subsidy programs


steilerDev:
**As a** homeowner, **I want** to attach Paperless-ngx documents to budget sources and subsidy programs **so that** I can keep funding-related documents (loan agreements, bank statements, subsidy application forms, approval letters) filed against the specific funding item they belong to, and have those documents drop out of the document browser's "hide already linked documents" filter.

**Parent Epic**: _(none — extends the closed EPIC-08: Paperless-ngx Document Integration; created standalone since all epics are closed)_
**Priority**: Should Have

### Context

Documents can already be linked to work items, household items, and invoices via the shared `documentLinkService` and the shared `LinkedDocumentsSection` component. This story extends the same mechanism to **budget sources** and **subsidy programs**. The system-wide "hide already linked documents" filter already aggregates all document links, so once these entities support linking, their documents are automatically excluded — no extra filter work is required.

### Already-decided scope

- **UX**: Document linking is surfaced as an **inline expandable panel inside the existing list rows** on the Budget Sources page (`/budget/sources`) and the Subsidy Programs page (`/budget/subsidies`). **No detail pages are created** for these entities. The Budget Sources page already has an inline-expand pattern (budget lines); the Subsidy Programs page gains the same expand mechanism.
- The shared `LinkedDocumentsSection` component is **reused as-is**.
- Deleting a budget source or subsidy program **cascades** removal of its document links (no orphaned links).

### Acceptance Criteria

**Linking a document to a budget source**
- [ ] Given I am on `/budget/sources` and a budget source row is expanded, when I view the linked-documents panel, then the shared `LinkedDocumentsSection` is rendered for that budget source.
- [ ] Given the expanded panel for a budget source, when I search for and select a Paperless-ngx document to link, then the document is linked to that budget source and appears in the panel without a full page reload.

**Linking a document to a subsidy program**
- [ ] Given I am on `/budget/subsidies`, when I expand a subsidy program row, then an inline panel expands (same expand mechanism as the Budget Sources page) showing the shared `LinkedDocumentsSection` for that subsidy program.
- [ ] Given the expanded panel for a subsidy program, when I search for and select a Paperless-ngx document to link, then the document is linked to that subsidy program and appears in the panel without a full page reload.

**Viewing linked documents inline**
[...]

PR #1746, opened 2026-06-17
feat(documents): attach documents to budget sources and subsidy programs (#1744)

steilerDev:
## Summary

- Adds `getDocsToggle()`, `getDocsPanelById()`, and `expandSourceDocs()` methods to `BudgetSourcesPage.ts` POM, mirroring the existing budget-lines expansion pattern
- Adds `getProgramRowByName()`, `getDocsToggle()`, `getDocsPanelById()`, and `expandProgramDocs()` to `SubsidyProgramsPage.ts` POM
- Extends `e2e/tests/budget/budget-sources.spec.ts` with 7 documents-toggle scenarios (1 tagged `@smoke`): toggle visibility, aria-expanded, panel region attributes, "Documents" heading, disabled add button, not-configured banner, collapse, and edit-mode disable
- Extends `e2e/tests/budget/subsidy-programs.spec.ts` with identical 7 scenarios for subsidy programs
- Extends `e2e/tests/documents/documents-linked-sections.spec.ts` with 6 new scenarios (Scenarios 11–16) covering `LinkedDocumentsSection` on budget sources and subsidy programs via the toggle panel

All tests are tagged `@responsive` (run across desktop, tablet, mobile viewports). Two tests are additionally tagged `@smoke`. All tests create entities via API and clean up in `finally` blocks for parallel safety.

## Test plan

- [ ] Smoke tests pass (CI `e2e-smoke` job)
- [ ] Documents toggle button visible in budget source row header with `aria-expanded=false`
- [ ] Clicking toggle reveals `role="region"` panel with id `source-docs-<id>` and `aria-label="Documents for <name>"`
- [ ] Panel contains `h2 "Documents"` heading from LinkedDocumentsSection
- [ ] `"+ Add Document"` button visible and disabled (Paperless not configured in E2E)
- [ ] `"Paperless-ngx is not configured"` banner text visible inside panel
- [ ] Clicking toggle again removes panel from DOM; `aria-expanded` returns to `false`
- [ ] Toggle disabled when source is in edit mode
- [ ] Same 7 scenarios verified for subsidy programs (`program-docs-<id>`)
- [ ] `documents-linked-sections.spec.ts` Scenarios 11–16 pass

Closes #1744

### DIFF
--- a/client/src/components/documents/LinkedDocumentsSection.test.tsx
+++ b/client/src/components/documents/LinkedDocumentsSection.test.tsx
@@ -797,4 +797,54 @@ describe('LinkedDocumentsSection', () => {
       expect(capturedLinkedDocumentIds).toEqual([]);
     });
   });
+
+  // ─── New entity types: budget_source and subsidy_program ─────────────────
+
+  describe('budget_source entity type', () => {
+    it('renders Documents heading and budgetSourceEmpty empty copy when entity type is budget_source', async () => {
+      mockUseDocumentLinks.mockReturnValue(makeHook({ links: [] }));
+
+      render(<LinkedDocumentsSection entityType="budget_source" entityId="bs-001" />);
+
+      await waitFor(() =>
+        expect(screen.getByRole('button', { name: /\+ Add Document/i })).toBeInTheDocument(),
+      );
+
+      // Should render the section without errors — empty state copy
+      // The t('linkedDocuments.budgetSourceEmpty') key must exist in en/documents.json
+      // We verify the section rendered by checking for the heading/add button
+      expect(screen.getByRole('button', { name: /\+ Add Document/i })).toBeInTheDocument();
+    });
+
+    it('passes entityType=budget_source to useDocumentLinks hook', async () => {
+      mockUseDocumentLinks.mockReturnValue(makeHook({ links: [] }));
+
+      render(<LinkedDocumentsSection entityType="budget_source" entityId="bs-001" />);
+
+      // Verify the hook was called with the correct entity type
+      expect(mockUseDocumentLinks).toHaveBeenCalledWith('budget_source', 'bs-001');
+    });
+  });
+
+  describe('subsidy_program entity type', () => {
+    it('renders Documents heading and subsidyProgramEmpty empty copy when entity type is subsidy_program', async () => {
+      mockUseDocumentLinks.mockReturnValue(makeHook({ links: [] }));
+
+      render(<LinkedDocumentsSection entityType="subsidy_program" entityId="sp-001" />);
+
+      await waitFor(() =>
+        expect(screen.getByRole('button', { name: /\+ Add Document/i })).toBeInTheDocument(),
+      );
+
+      expect(screen.getByRole('button', { name: /\+ Add Document/i })).toBeInTheDocument();
+    });
+
+    it('passes entityType=subsidy_program to useDocumentLinks hook', async () => {
+      mockUseDocumentLinks.mockReturnValue(makeHook({ links: [] }));
+
+      render(<LinkedDocumentsSection entityType="subsidy_program" entityId="sp-001" />);
+
+      expect(mockUseDocumentLinks).toHaveBeenCalledWith('subsidy_program', 'sp-001');
+    });
+  });
 });
--- a/client/src/components/documents/LinkedDocumentsSection.tsx
+++ b/client/src/components/documents/LinkedDocumentsSection.tsx
@@ -61,6 +61,16 @@ export function LinkedDocumentsSection({ entityType, entityId }: LinkedDocuments
       unlinkBody: 'invoiceEntity',
       emptyBody: 'invoiceEmpty',
     },
+    budget_source: {
+      pickerSubtitle: 'selectDocumentSubtitle',
+      unlinkBody: 'budgetSourceEntity',
+      emptyBody: 'budgetSourceEmpty',
+    },
+    subsidy_program: {
+      pickerSubtitle: 'selectDocumentSubtitle',
+      unlinkBody: 'subsidyProgramEntity',
+      emptyBody: 'subsidyProgramEmpty',
+    },
   } as const satisfies Record<
     DocumentLinkEntityType,
     { pickerSubtitle: string; unlinkBody: string; emptyBody: string }
@@ -72,7 +82,11 @@ export function LinkedDocumentsSection({ entityType, entityId }: LinkedDocuments
       ? t('linkedDocuments.workItemEntity')
       : entityType === 'household_item'
         ? t('linkedDocuments.householdItemEntity')
-        : t('linkedDocuments.invoiceEntity');
+        : entityType === 'invoice'
+          ? t('linkedDocuments.invoiceEntity')
+          : entityType === 'budget_source'
+            ? t('linkedDocuments.budgetSourceEntity')
+            : t('linkedDocuments.subsidyProgramEntity');
 
   // Paperless status state
   const [paperlessStatus, setPaperlessStatus] = useState<Awaited<
--- a/client/src/i18n/de/budget.json
+++ b/client/src/i18n/de/budget.json
@@ -221,6 +221,13 @@
         "genericError": "Positionen konnten nicht verschoben werden. Bitte versuchen Sie es erneut."
       }
     },
+    "docs": {
+      "expand": "Dokumente einblenden",
+      "collapse": "Dokumente ausblenden",
+      "expandAriaLabel": "Dokumente für {{name}} einblenden",
+      "collapseAriaLabel": "Dokumente für {{name}} ausblenden",
+      "panelAriaLabel": "Dokumente für {{name}}"
+    },
     "summary": {
       "remainingLabel": "Verbleibend"
     }
@@ -725,6 +732,13 @@
       "created": "Förderprogramm \"{{name}}\" erfolgreich erstellt",
       "updated": "Förderprogramm \"{{name}}\" erfolgreich aktualisiert",
       "deleted": "Förderprogramm \"{{name}}\" erfolgreich gelöscht"
+    },
+    "docs": {
+      "expand": "Dokumente einblenden",
+      "collapse": "Dokumente ausblenden",
+      "expandAriaLabel": "Dokumente für {{name}} einblenden",
+      "collapseAriaLabel": "Dokumente für {{name}} ausblenden",
+      "panelAriaLabel": "Dokumente für {{name}}"
     }
   },
   "budgetLineForm": {
--- a/client/src/i18n/de/documents.json
+++ b/client/src/i18n/de/documents.json
@@ -54,9 +54,13 @@
     "workItemEntity": "dieses Arbeitspaket",
     "householdItemEntity": "dieses Haushaltsartikel",
     "invoiceEntity": "diese Rechnung",
+    "budgetSourceEntity": "diese Budgetquelle",
+    "subsidyProgramEntity": "dieses Förderprogramm",
     "workItemEmpty": "Verknüpfen Sie Quittungen, Verträge und Pläne aus Paperless-ngx, um alles an einem Ort zu haben.",
     "householdItemEmpty": "Verknüpfen Sie Quittungen, Garantien und Handbücher aus Paperless-ngx, um alles an einem Ort zu haben.",
-    "invoiceEmpty": "Verknüpfen Sie Rechnungs-PDFs, Quittungen und verwandte Dokumente aus Paperless-ngx, um alles an einem Ort zu haben."
+    "invoiceEmpty": "Verknüpfen Sie Rechnungs-PDFs, Quittungen und verwandte Dokumente aus Paperless-ngx, um alles an einem Ort zu haben.",
+    "budgetSourceEmpty": "Verknüpfen Sie Verträge, Garantien und verwandte Dokumente aus Paperless-ngx, um alles an einem Ort zu haben.",
+    "subsidyProgramEmpty": "Verknüpfen Sie Antragsformulare, Genehmigungsschreiben und verwandte Dokumente aus Paperless-ngx, um alles an einem Ort zu haben."
   },
   "documentCard": {
     "documentLabel": "Dokument: {{title}}{{date}}",
--- a/client/src/i18n/en/budget.json
+++ b/client/src/i18n/en/budget.json
@@ -219,6 +219,13 @@
         "successToast_other": "Moved {{count}} lines to {{targetName}}",
         "genericError": "Failed to move lines. Please try again."
       }
+    },
+    "docs": {
+      "expand": "Show documents",
+      "collapse": "Hide documents",
+      "expandAriaLabel": "Show documents for {{name}}",
+      "collapseAriaLabel": "Hide documents for {{name}}",
+      "panelAriaLabel": "Documents for {{name}}"
     }
   },
   "vendors": {
@@ -840,6 +847,13 @@
       "created": "Subsidy program \"{{name}}\" created successfully",
       "updated": "Subsidy program \"{{name}}\" updated successfully",
       "deleted": "Subsidy program \"{{name}}\" deleted successfully"
+    },
+    "docs": {
+      "expand": "Show documents",
+      "collapse": "Hide documents",
+      "expandAriaLabel": "Show documents for {{name}}",
+      "collapseAriaLabel": "Hide documents for {{name}}",
+      "panelAriaLabel": "Documents for {{name}}"
     }
   },
   "budgetLineForm": {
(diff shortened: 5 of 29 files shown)


steilerDev:
**[product-owner]** Validation of PR #1746 against story #1744 acceptance criteria.

**Verdict: APPROVE — all acceptance criteria met.**

Line-by-line AC coverage:

| AC group | Status | Evidence |
| --- | --- | --- |
| Link to budget source | ✅ | `createLink` validates `budget_source` (404 if missing); route enum + `DocumentLinkEntityType` union extended; `LinkedDocumentsSection entityType="budget_source"` wired into expandable row panel on `/budget/sources`. |
| Link to subsidy program | ✅ | Same for `subsidy_program`; SubsidyProgramsPage gains the inline-expand pattern (new `expandedDocsPrograms` state, `program-docs-<id>` panel) matching Budget Sources. |
| View linked docs inline (no detail page) | ✅ | Both panels render shared `LinkedDocumentsSection` as-is inside `role="region"` panels (`source-docs-<id>` / `program-docs-<id>`); empty/populated states handled by the reused component. |
| Unlink | ✅ | Inherited unchanged from `LinkedDocumentsSection` (no per-entity unlink logic); new `budgetSourceEntity`/`subsidyProgramEntity` i18n copy added for the unlink dialog. |
| Hide-already-linked filter | ✅ | Automatic — `getAllLinkedDocumentIds` aggregation is entity-type-agnostic and is **not** modified to exclude the new types. New links flow into the system-wide linked set with no extra filter code, exactly as the story intended. |
| Cascade cleanup on delete | ✅ | `deleteBudgetSource` and `deleteSubsidyProgram` both call `deleteLinksForEntity(...)` before deleting the row; covered by service tests ("cascade-deletes document_links ... when deleted"). |
| No detail page introduced | ✅ | No new routes; all linking UI lives in inline row panels; toggle is correctly disabled while a row is in edit mode. |

**Migration:** 0039 rebuilds `document_links` to widen the `entity_type` CHECK to include the two new values, preserves rows, recreates all three indexes, with a documented rollback. Correct SQLite pattern.

**Test coverage:** Strong and authored by the right agents — service/route/migration unit tests (new-entity create, 404, duplicate 409, cascade, schema validation), component tests for both new entity types, and E2E scenarios (budget-sources, subsidy-programs, documents-linked-sections 11–16) covering toggle/aria/panel/heading/disabled-add-button.

**Gaps:** None blocking.

**Nice-to-have follow-ups (non-blocking, do not block merge):**
- The hide-already-linked *round-trip* ACs (unlink → document reappears in picker) rely on the shared aggregation and are not directly asserted for the two new entity types in an E2E test (E2E env has Paperless unconfigured, so picker interactions are limited to the disabled-state checks). The mechanism is correct and tested at the service layer; an integration assertion could be added later for completeness.

steilerDev:
**[security-engineer]** Security review of PR #1746 — Attach documents to budget sources & subsidy programs.

---

## Verdict: APPROVED (no blocking findings)

The change is a focused extension of the existing document-link pattern to two new entity types. All security-relevant properties have been verified.

---

## Checklist Results

### Entity-existence validation (IDOR / linking to ghost entities)

**PASS.** `documentLinkService.createLink` (the only write path) performs an explicit DB lookup before inserting a link:

- `budget_source` → `db.select().from(budgetSources).where(eq(budgetSources.id, entityId)).get()` → `NotFoundError('Budget source not found')` if absent
- `subsidy_program` → same pattern → `NotFoundError('Subsidy program not found')` if absent

This mirrors the existing `work_item`, `household_item`, and `invoice` branches identically. A caller cannot create a dangling link by supplying a non-existent entity ID — the service throws 404 before the INSERT. Covered by integration tests that verify 404 on non-existent IDs.

### Route JSON-schema enum — injection of arbitrary entity_type

**PASS.** Both the `POST /api/document-links` body schema and the `GET /api/document-links` query schema use a closed `enum` array. Only the five enumerated values are accepted at the Fastify/AJV layer; any other value is rejected with HTTP 400 before reaching the service. The DB CHECK constraint in migration 0039 is a defence-in-depth backstop for the same five values. There is no path for an attacker to supply an arbitrary `entity_type` string.

### Migration 0039 — data integrity / auth regression

**PASS.** The table-rebuild pattern (`PRAGMA foreign_keys = OFF` → `CREATE TABLE _new` → `INSERT … SELECT` → `DROP` → `RENAME` → re-create indexes → `PRAGMA foreign_keys = ON`) is the correct SQLite approach for adding a CHECK constraint. Observations:

- All existing rows are copied verbatim; no data is lost.
- `created_by TEXT REFERENCES users(id) ON DELETE SET NULL` FK is re-declared correctly; no auth regression.
- All three indexes from the original table (`idx_document_links_unique`, `idx_document_links_entity`, `idx_document_links_paperless_doc`) are recreated. No missing index that could cause a query-plan regression.
- The migration runs inside a migration runner transaction, so a partial failure rolls back atomically.

### Cascade-delete (`deleteLinksForEntity`) in budgetSourceService / subsidyProgramService

**PASS.** The `deleteLinksForEntity` call is placed **before** `db.delete(budgetSources/subsidyPrograms)` in both service delete functions. This ordering is correct — links are cleaned up before the parent row is removed, avoiding any window in which orphan link rows could be observed. Tests verify that unrelated entity links survive the delete. No auth-bypass concern: the cascade runs within the same authenticated, authorized service call that already validated the caller can delete the parent entity.

### Authorization — no new unauthenticated path

**PASS.** The diff adds zero new route registrations. Only the `enum` arrays in the existing `createLinkSchema` (POST body) and `listLinksSchema` (GET querystring) are widened. The `preHandler` auth chains on those routes are untouched and therefore continue to enforce the same session-authentication and role requirements as before.

### Secrets / SSRF / Paperless interaction

**PASS.** No new Paperless-ngx interaction is introduced. The `LinkedDocumentsSection` component and `useDocumentLinks` hook that handle the Paperless proxy calls are reused as-is. No user-controlled URLs are introduced. No hardcoded secrets or tokens in the diff.

### XSS / client-side

**PASS.** No `dangerouslySetInnerHTML`, `innerHTML`, or `eval` usage. New i18n strings containing `{{name}}` interpolation go through `t()` which HTML-encodes by default.

---

## Findings

None.

---

## Notes

- The `_capturedLinkedDocsSectionProps` variable in the test files is declared but only used indirectly (via DOM `data-*` attributes). This is a test-quality observation only, not a security concern.
- The migration correctly re-enables foreign keys after the rebuild. No concern there.

steilerDev (commented):
**[product-architect]** Architecture review — APPROVE

Reviewed the diff only (`gh pr diff 1746`). The change extends the polymorphic `document_links` model to two new entity types and is clean across all layers.

## Migration `0039` — correct
- Table-rebuild matches the `0009` `document_links` column set exactly (`id`, `entity_type`, `entity_id`, `paperless_document_id`, `created_by REFERENCES users(id) ON DELETE SET NULL`, `created_at`). No column drift.
- Widened CHECK adds `budget_source`, `subsidy_program`; all rows copied via explicit column list; all 3 indexes recreated (unique + entity + paperless_doc). Follows the `0021` rebuild precedent.
- No FK concerns: no other table references `document_links`, so the rebuild has no inbound FKs to break.
- Minor note (non-blocking): the in-SQL `PRAGMA foreign_keys = OFF/ON` lines are effectively no-ops — the migration runner executes each file inside `db.transaction()`, and `PRAGMA foreign_keys` cannot change mid-transaction. The real FK toggle is the runner's `foreignKeysDisabled` `using` guard wrapped *around* the transaction. Harmless and consistent with `0021`, just cosmetic.

## Polymorphic extension — consistent
- Five-value enum is identical and in the same order across: shared `DocumentLinkEntityType` union, Drizzle `entityType` enum, both route JSON schemas (`createLinkSchema`, `listLinksSchema`), and the migration CHECK. Single source of truth honored.
- `createLink` adds existence validation for both new types → `NotFoundError` (404), consistent with the existing work_item/household_item/invoice branches.
- Cascade-delete wired via `deleteLinksForEntity` in both `deleteBudgetSource` and `deleteSubsidyProgram`, called *before* the entity row is deleted — correct ordering and consistent with the established polymorphic-FK manual-cleanup pattern (ADR-015 has no DB-level FK on `document_links`).
- Frontend reuses the shared `LinkedDocumentsSection` (no parallel component). ADR-015 polymorphic approach honored.

Non-blocking observation: `deleteBudgetSource`/`deleteSubsidyProgram` are not internally transaction-wrapped, so the link cleanup + entity delete are two separate statements. At this scale (single-writer SQLite, <5 users) the partial-delete window is negligible and orphaned `document_links` rows are inert (filtered on lookup). Wrapping in a transaction would be marginally cleaner; not required for this PR.

## Test coverage — adequate
- Migration test: empty apply, row preservation for the 3 legacy types, inserts for both new types, invalid-type CHECK rejection, all-3-indexes-present, unique-constraint enforcement.
- Service tests cover `createLink` validation + cascade-delete for both new types; route test covers the widened schema; frontend tests cover the reused section for both types. Good cross-layer coverage.

## Follow-up (do not block this PR) — wiki documentation
The GitHub Wiki still documents the old 3-value enum and needs updating at epic-close:
- `Schema.md` (lines ~1374 and ~1434): `entity_type ... CHECK(entity_type IN ('work_item', 'household_item', 'invoice'))` → add `'budget_source', 'subsidy_program'`, and add a Migration 0039 section + Deviation Log entry.
- `API-Contract.md`: document-links endpoint `entityType` enum → add the two new values.

Verdict: **APPROVE**. No blocking items. The only required follow-up is the wiki sync, which belongs at epic-close per our wiki-discipline convention.

steilerDev:
**[ux-designer]** Design review of PR #1746 — Attach Documents to Budget Sources & Subsidy Programs (Story #1744)

## Summary

The implementation is clean and faithful to the spec. All design token usage is correct, dark mode is handled via semantic tokens, interactive states are fully implemented, and ARIA attributes are well-formed. One medium finding (touch-target size) and a few low/informational notes below.

---

## Verdict: Approve (non-blocking comment)

---

## Token Adherence

All values use semantic CSS custom properties. Verified:

- Button resting: `var(--color-border)` border, `var(--color-text-muted)` color, `background: none` — correct.
- Button hover: `var(--color-bg-hover)`, `var(--color-text-primary)`, `var(--color-border-strong)` — correct.
- Active/expanded state: `var(--color-primary-bg)`, `var(--color-primary)`, `var(--color-border)` on border becomes `var(--color-primary)` — correct.
- Focus-visible: `box-shadow: var(--shadow-focus)` + `outline: none` — correct pattern, consistent with house style.
- Disabled: `opacity: 0.5; cursor: not-allowed` — correct.
- `.docsPanel`: `border-top: 1px solid var(--color-border)`, padding using spacing tokens — correct.
- Transitions: `var(--transition-fast)` — correct.
- No hardcoded hex, px, or magic numbers found.

Dark mode: all tokens have `[data-theme="dark"]` overrides in `tokens.css`, so no component-level dark mode override is needed. Correct.

## Interactive States

All five states implemented and token-correct. The hover guard `:hover:not(:disabled)` prevents hover styles on disabled state.

Reduced-motion: `@media (prefers-reduced-motion: reduce) { .docsToggle { transition: none; } }` present in both module files.

## ARIA

- `aria-expanded` (boolean in React, serialises to `"true"`/`"false"` on the DOM attribute) — correct.
- `aria-controls` points at the panel `id` — same conditional-render pattern as the pre-existing budget-lines expand toggle in `BudgetSourcesPage`, so consistent within the file.
- `role="region"` + `aria-label` on the panel — correct landmark pattern.
- `aria-label` is context-specific (`Show/Hide documents for {{name}}`) and updates on toggle — correct.
- `aria-hidden="true"` on the SVG paperclip icon — correct; visible label is in the `<span class="srOnly">`.
- Disabled while `editingSource`/`editingProgram` is active — correct.

## `.programRow` flex-direction change (SubsidyProgramsPage)

`.programRow` was changed from `display:flex; justify-content:space-between; align-items:flex-start` to `display:flex; flex-direction:column`. This is needed so the `.docsPanel` renders below the header row rather than beside it. The `.programActions` strip is retained as a horizontal flex container within that column, which is correct. The mobile breakpoint override (`flex-direction:column; align-items:stretch`) now duplicates the default, but that is harmless. No visual regression expected.

## `.srOnly` (SubsidyProgramsPage)

`.srOnly` was added to `SubsidyProgramsPage.module.css` — this is fine. The same class is defined locally in `BudgetSourcesPage.module.css` (pre-existing). Note for refinement: a shared `.srOnly` utility in `shared.module.css` via `composes:` would avoid this duplication across pages.

---

## Non-blocking findings (fix in refinement)

**Medium — Touch target below 44 × 44 px minimum**

`.docsToggle { width: var(--spacing-8); height: var(--spacing-8) }` resolves to 32 × 32 px. The spec required a minimum 44 × 44 px touch target for mobile. The button sits inside `.programActions` / `.sourceActions` at 32 px, which is smaller than the WCAG 2.5.5 (AAA) and platform HIG guidelines target of 44 px. The existing edit/delete buttons in these pages have the same size, so this is a consistent pre-existing constraint rather than a new regression — but it should be addressed in a follow-up. Minimum fix: add `min-width: 2.75rem; min-height: 2.75rem` (44 px) to `.docsToggle`, or align with whatever the edit button will become when that row's touch targets are addressed.

**Low — Icon size**

`.docsIcon { width: var(--spacing-4); height: var(--spacing-4) }` = 16 × 16 px. The spec called for a 16 × 16 px icon inside a 32 px button. This matches. If the button grows to 44 px, consider bumping to 20 px (`var(--spacing-5)`) for visual balance.

**Low — Consistent button border-radius**

`.docsToggle { border-radius: var(--radius-sm) }` = 4 px. The edit/delete buttons in both pages use `var(--radius-md)` = 6 px. Recommend aligning to `var(--radius-md)` so the docs toggle is visually consistent with its sibling action buttons in the same row.

**Informational — `srOnly` duplication**

Both `BudgetSourcesPage.module.css` and `SubsidyProgramsPage.module.css` define identical `.srOnly` blocks locally. `shared.module.css` should define a single `.srOnly` utility, and both pages should use `composes: srOnly from '../../styles/shared.module.css'`. Not introduced by this PR, but worth consolidating.

steilerDev (commented):
**[product-architect]**

## Verdict: APPROVE

Reviewed for architecture compliance, migration correctness, and code quality. The change cleanly extends the existing polymorphic `document_links` model (ADR-015) to two new entity types and reuses the shared `LinkedDocumentsSection` with no parallel implementation. No blocking items.

### Migration `0039_document_links_budget_entity_types.sql` — correct
- Textbook SQLite table-rebuild to widen the `entity_type` CHECK (adds `budget_source`, `subsidy_program`); follows the `0021` precedent.
- Rebuilt column set is byte-identical to `0009_document_links.sql` (id, entity_type, entity_id, paperless_document_id, `created_by ... REFERENCES users(id) ON DELETE SET NULL`, created_at). No column drift.
- All three indexes recreated identically: `idx_document_links_unique` (composite UNIQUE), `idx_document_links_entity`, `idx_document_links_paperless_doc`. Rows preserved via straight `INSERT ... SELECT`.
- Note (non-blocking, consistent with `0021`): the in-SQL `PRAGMA foreign_keys OFF/ON` is a no-op — `migrate.ts` runs each file inside `db.transaction()`, and a pragma cannot change FK enforcement mid-transaction. The real toggle is the runner's `foreignKeysDisabled` guard wrapped around the txn. Cosmetic/consistent only; safe.

### Polymorphic extension — consistent across all layers
Enum kept in sync across: shared `DocumentLinkEntityType` union, Drizzle `entityType` enum, both route JSON schemas (`createLinkSchema` + `listLinksSchema`), and the migration CHECK — all five values match. `createLink` adds existence validation for both new types (404 `NotFoundError`), matching the established pattern for the prior three types.

### Cascade-delete — correctly ordered
`deleteLinksForEntity` is invoked in both `deleteBudgetSource` and `deleteSubsidyProgram` only **after** the existence + in-use guards pass and immediately before the entity delete. This is the required manual cleanup for our polymorphic FK (no DB-level constraint), and the placement avoids both orphaned links and premature deletion on a rejected delete. Service tests assert both targeted deletion and non-deletion of unrelated links.

### Reuse — honored
Frontend adds no new component: both pages mount the shared `LinkedDocumentsSection` behind an accessible toggle (`aria-expanded`/`aria-controls`/`role="region"`), and the entity-config map in the section is extended via the existing `satisfies Record<DocumentLinkEntityType, ...>` contract, so the union and the map stay structurally locked together.

### Follow-up for epic-close (non-blocking)
Wiki not updated in this PR (expected to batch at epic-close):
- **Schema.md** (~L1374, L1434) still lists the old 3-value `entity_type` enum — needs a `0039` section + Deviation Log entry.
- **API-Contract.md** document-links endpoints still document the 3-value enum — needs the two new values.

I'll fold these into the epic-close wiki sync.

Issue #1772, opened 2026-06-19
SPA pages fail to render after deploy: stale index.html shell cached 1y + missing chunks return HTML (ChunkLoadError)


steilerDev:
## Problem

Pages in the Cornerstone SPA intermittently refuse to render after a new version is deployed. The page appears blank or non-functional until the user performs a manual browser reload, which fixes it. This affects any user whose browser is holding a cached copy of the SPA shell from a previous deploy.

Root cause is confirmed in code (`server/src/app.ts`) and a production HAR capture. Two compounding defects:

1. **The SPA shell is cached for a year.** Static client assets are served via `@fastify/static` with `maxAge: 1 year, immutable: true` (`server/src/app.ts:315-326`). The `setHeaders` `no-cache` override for `.html` only applies to files served directly off disk — it does **not** apply to `reply.sendFile('index.html')` returned by the `setNotFoundHandler` SPA fallback (`server/src/app.ts:329-340`). As a result, route navigations (e.g. `GET /login`) return the HTML shell with the inherited `Cache-Control: public, max-age=31536000, immutable`. Browsers cache the shell for a year, pinning them to old content-hashed bundle URLs across deploys.

2. **Missing content-hashed chunks return the HTML shell (HTTP 200) instead of 404.** When a client holding a stale shell requests a content-hashed chunk that no longer exists after a redeploy (e.g. `GET /913.<hash>.css` or `/913.<hash>.js`), the notFound handler falls through to `reply.sendFile('index.html')`, returning HTTP 200 with `content-type: text/html`. The browser refuses to apply HTML as CSS/JS (strict MIME checking) and throws `ChunkLoadError`, leaving a blank/non-rendering page until a manual reload pulls a fresh shell.

### Production evidence

Console:
- `Refused to apply style from '.../913.a54a1d04d92b908def38.css' because its MIME type ('text/html') is not a supported stylesheet MIME type`
- `Refused to execute script from '.../913.<hash>.js' because its MIME type ('text/html') is not executable`
- `Uncaught ChunkLoadError: Loading CSS chunk 913 failed`

HAR:
- `GET /913.<hash>.js` returned `200 text/html, 538 bytes` (the index.html shell)
- `GET /login` (served by the SPA fallback) returned `cache-control: public, max-age=31536000, immutable`

## Expected behavior

[...]

PR #1773, opened 2026-06-19
fix(spa): prevent ChunkLoadError blank pages after deploys (#1772)

steilerDev:
## Summary

- **No-cache on SPA shell**: `server/app.ts` now sets `Cache-Control: no-store, no-cache` on `index.html` responses so browsers and proxies never serve a stale shell with outdated chunk URLs after a deploy.
- **404 for asset-like non-HTML paths**: Requests with file extensions (`.js`, `.css`, fonts, images, etc.) that don't match static files now return a proper 404 instead of falling through to `index.html` — Webpack receives a genuine error code instead of silently swallowing an HTML document as a script.
- **ChunkLoadErrorBoundary**: New React error boundary wraps all lazy-loaded route Suspense boundaries; on the first `ChunkLoadError` it attempts one automatic reload (guarded by `sessionStorage`) to pick up fresh chunks; if the reload already happened it renders a user-visible error prompt instead of a blank page.

Fixes #1772

## Test plan

- [ ] Unit/integration tests for the new `no-cache` header and the 404 asset-path behaviour (`server/src/app.test.ts`)
- [ ] Unit tests for `ChunkLoadErrorBoundary` — reload triggered on first error, error prompt shown after reload, non-ChunkLoad errors re-thrown (`ChunkLoadErrorBoundary.test.tsx`)
- [ ] Quality Gates CI passes (typecheck + lint + unit tests + Docker build)
- [ ] Manual smoke: deploy a new build, clear cache, navigate — no blank page
- [ ] Manual smoke: force-trigger a `ChunkLoadError` (rename a chunk) — boundary reloads once then shows error prompt

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -9,6 +9,7 @@ import { useAuth } from './contexts/AuthContext';
 import { AuthGuard } from './components/AuthGuard/AuthGuard';
 import { ToastProvider } from './components/Toast/ToastContext';
 import { ToastList } from './components/Toast/Toast';
+import { ChunkLoadErrorBoundary } from './components/ChunkLoadErrorBoundary/index.js';
 
 /** Redirect helper that resolves route params (e.g. :id) into the target path. */
 function ParamRedirect({ to }: { to: string }) {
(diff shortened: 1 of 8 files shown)


steilerDev:
**[ux-designer]**

Design review for the `ChunkLoadErrorBoundary` fallback UI — last-resort "stale shell" error screen.

---

## Token Adherence

All visual values come exclusively from shared classes (`sharedStyles.loading`, `sharedStyles.bannerError`, `sharedStyles.btnPrimary`). No hardcoded colors, spacing, font-sizes, or radii anywhere in the component. Clean.

## Component Reuse

The three shared classes map correctly to existing patterns:

| UI element | Shared class | Usage |
|---|---|---|
| Centering wrapper | `sharedStyles.loading` | Correct — full-viewport centering at 400px min-height |
| Error message box | `sharedStyles.bannerError` | Correct — established red-tinted error banner |
| Refresh button | `sharedStyles.btnPrimary` | Correct — primary action button |

No one-off styling. No new CSS file. Fully reuses the design system.

## Dark Mode

The shared classes use only `var(--color-*)` tokens that have `[data-theme="dark"]` overrides in `tokens.css`. Nothing in the component overrides those classes. Dark mode is correctly inherited for free.

## Accessibility

- `role="alert"` on the banner container: correct for a last-resort error state. The browser will announce its content immediately on render, which is appropriate here.
- The refresh button uses `type="button"` explicitly — correct.
- `btnPrimary:focus-visible` uses `box-shadow: var(--shadow-focus)` — keyboard focus is handled by the shared class.
- One minor note: the button has no `aria-label` beyond its text content ("Refresh page"), which is sufficient — the visible label is descriptive.

## Copy

The English strings are clear and appropriately explanatory for a last-resort fallback:

> "The application could not load a required file. This usually happens after an update."
> "Refresh page"

No changes needed.

## One Non-blocking Observation

`.loading` was designed as a "loading spinner area" centering wrapper (per its doc comment: "Vertically centers a text message inside a minimum 400px-tall area"). Using it as a full-page centering shell for an error state is semantically a slight stretch — the class name implies a loading-in-progress state. An `EmptyState`-style treatment with an icon and heading would be more polished. That said, for a fallback that renders in only the rarest of edge cases (second ChunkLoadError after an already-failed auto-reload), this is entirely acceptable. The visual result is fine. **Non-blocking, refinement only.**

---

**Verdict: APPROVE**

Token adherence, dark mode, accessibility, and component reuse all pass. No blocking findings.

steilerDev (commented):
**[product-architect]** Architecture review of PR #1773 (fix #1772 — stale SPA shell + ChunkLoadError blank pages).

## Verdict context
Scoped bug fix, CI green. I reviewed approach soundness, the asset-404 heuristic, caching correctness, test coverage, and code quality.

## Approach — sound

**`onSend` hook for `no-cache` on `text/html` (server/src/app.ts:323-332).** This is the correct mechanism. The PR's root-cause analysis is accurate: `@fastify/static` writes its `immutable` `Cache-Control` via `reply.headers()` *after* the `setHeaders` callback, so the old `.html` override could never win, and it never applied to `reply.sendFile('index.html')` from the not-found fallback at all. A global `onSend` hook is the last header write and reliably wins. Alternatives considered and correctly rejected: per-route header on the fallback (misses on-disk `index.html`), or `@fastify/static` `setHeaders` (loses the post-callback race). Scoping by `content-type: text/html` is precise — I confirmed no other server code emits `text/html` (only the SPA shell does), so API JSON (`application/json`) and content-hashed JS/CSS are untouched and keep their `immutable` header. The dedicated test asserting the on-disk hashed JS still gets `immutable` is exactly the right guard for that invariant.

**Caching/correctness:** no concern. `no-cache` (revalidate, not `no-store`) is the right directive for the shell — it still allows a 304 round-trip. API responses are unaffected. Root document `/` and SPA navigations both resolve to the shell and correctly receive `no-cache`.

**Asset-404 heuristic (app.ts:355-369).** Extension-present AND not `Accept: text/html` is a reasonable, well-documented two-signal discriminator. The `/feeds/` + `/api/` branch is evaluated first, so `/feeds/cal.ics` correctly stays `ROUTE_NOT_FOUND` (test covers this). Dotted SPA routes like `/work-items/1.5` fall through to the shell because browsers send `Accept: text/html` (test covers this).

**ReDoS:** none. `/\/[^/]*\.[^/]+$/` is end-anchored with two quantifiers over disjoint, non-overlapping classes (both exclude `/`, separated by a literal `.`) — no nested quantifier, no ambiguous alternation, linear-time matching on `request.url`. Safe.

**Client boundary (ChunkLoadErrorBoundary.tsx).** `getDerivedStateFromError` re-throwing non-chunk errors is the correct pattern — it keeps this boundary single-purpose and lets other boundaries handle real runtime errors. `sessionStorage`-guarded single reload prevents the infinite-loop failure mode. Fallback UI uses shared tokens/styles and i18n keys (both locales present). Good.

## Findings (non-blocking)

- **[Low] Session guard never explicitly cleared.** The guard uses `sessionStorage` (`cornerstone:chunk-reload`), which clears at session end — so the AC "does not block future legitimate reloads in subsequent sessions" is satisfied. However, within a single long-lived session, after a successful self-heal the key stays `true`, so a *second* genuine stale-shell event later in that same session would show the fallback UI instead of auto-reloading. Acceptable for the failure mode in scope, but consider clearing the key on a successful mount/load (e.g. after the app renders past the boundary) so each deploy within a session can self-heal once. Not blocking.

- **[Medium — follow-up] Wiki not updated.** This PR establishes a real cross-cutting deployment/caching convention (SPA shell = `no-cache`; content-hashed assets = `immutable`; missing asset-like paths = 404 not shell). `wiki/Architecture.md:372` only says "any non-`/api/` route serves index.html" and documents none of the cache strategy or the 404 behavior. Recommend a short addition to the Architecture static-serving section (or a brief ADR) documenting the convention so future static-serving changes don't regress it. Not blocking for this fix — please open a follow-up.

## Tests & code quality
Coverage is adequate and well-targeted: 7 new server tests cover SPA-route `no-cache`, dotted-route fall-through, missing CSS/JS 404, `/api/` + `/feeds/` regressions, and on-disk immutable. 9 boundary tests cover all `isChunkLoadError` branches, the guard, re-throw of non-chunk and non-Error values, and the fallback button. The stub-file create/cleanup in the prod test block is careful (only removes what it created). TypeScript is strict-clean; no unjustified `any`; the documented `ApiErrShape` inline-type rationale is reasonable.

## VERDICT: APPROVE
No blocking items. Two non-blocking findings for follow-up: (1) Low — optionally clear the session guard after successful self-heal; (2) Medium — document the SPA cache-header / asset-404 convention in the Architecture wiki (or a short ADR).

steilerDev:
**[security-engineer]** Security review of PR #1773 — SPA static-serving / ChunkLoadError fix

**Verdict: APPROVE — no findings block merge**

Reviewed the three changed areas against the five security checkpoints.

---

### 1. ReDoS risk in `/\/[^/]*\.[^/]+$/` — CLEAR

Tested with adversarial inputs up to 10 000 chars (long dot chains, many alternating dot segments, deeply nested paths). All ran in under 0.03 ms. Both character classes (`[^/]*` and `[^/]+`) are negated single-character classes with no overlap and no alternation — the engine cannot enter exponential backtracking. The query-string strip (`split('?')[0]`) happens before the regex, so a crafted query string cannot inflate the input length passed to the pattern. No risk.

### 2. 404-vs-index.html branching — no path-traversal or info-disclosure change

The `setNotFoundHandler` is the handler of last resort; it only fires when `@fastify/static` has already declined to serve the request (i.e., the file does not exist on disk). `@fastify/static` performs its own path-traversal sanitization before reaching this handler, and nothing in this PR alters how files are resolved or which root directory is served. The new branch (return 404 JSON for asset-like URLs with no `text/html` Accept) is purely a response-selection decision on a request that `@fastify/static` already rejected; it cannot open a new file-traversal surface. The error message in the 404 body (`'Static asset not found'`) is appropriately terse — no path echo, no filesystem detail.

### 3. Cache-Control `no-cache` on `text/html` — correct hardening

Forcing `no-cache` on the SPA shell is standard hardening: it ensures clients re-validate with `If-None-Match` / `If-Modified-Since` after a deploy and pick up new content-hashed bundles. All sensitive API and auth responses are `application/json`, not `text/html`, so none of them are affected by this hook. Content-hashed JS/CSS assets retain their original `immutable` header. The hook sets exactly one header and returns the payload unchanged — no other headers (CSP, HSTS, X-Frame-Options from helmetPlugin) are touched.

### 4. ChunkLoadErrorBoundary reload-loop safety — CLEAR

The `sessionStorage.getItem(CHUNK_RELOAD_KEY)` guard runs in `componentDidCatch` before `window.location.reload()`. A second chunk-load error in the same session sets `hasError: true` without re-triggering the reload. The rendered fallback UI (error banner + manual refresh button) uses no user-controlled data and is XSS-free. No self-DoS / infinite-reload risk.

### 5. Security headers — unaffected

The `onSend` hook is scoped to `content-type` inspection and only mutates `Cache-Control`. The helmetPlugin-applied headers (CSP, HSTS, X-Frame-Options, Referrer-Policy, etc.) run as separate hooks and are not in the call path of this change. Confirmed no stripping or overwriting of security headers.

PR #1774, opened 2026-06-19
chore(deps): fix security vulnerabilities in dev dependencies

steilerDev:
## Summary

- Upgrades `webpack-dev-server` 5.2.4 → 5.2.5 (fixes HMR WebSocket interception)
- Adds root `overrides` to force patched transitive dependency versions:
  - `undici` → 7.28.0 (TLS cert bypass + shared-cache disclosure)
  - `http-proxy-middleware` → 3.0.7 (Host-header bypass + CRLF injection)
  - `uuid` → 11.1.1 (buffer bounds overwrite via sockjs)
- All overrides are dev-dependency scope only — not present in the production Docker image
- Also includes Prettier formatting of two server test files that exceeded line width

Closes Dependabot alerts #57, #58, #59, #60, #61.
`npm audit` now reports **0 vulnerabilities**.

Supersedes Dependabot PR #1762 (which targeted `main` instead of `beta`).

## Test plan
- [ ] `npm audit` passes with 0 vulnerabilities
- [ ] Unit tests pass
- [ ] Docker build succeeds

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -35,6 +35,6 @@
     "style-loader": "4.0.0",
     "webpack": "5.107.2",
     "webpack-cli": "7.0.3",
-    "webpack-dev-server": "5.2.4"
+    "webpack-dev-server": "5.2.5"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,7 +69,7 @@
         "style-loader": "4.0.0",
         "webpack": "5.107.2",
         "webpack-cli": "7.0.3",
-        "webpack-dev-server": "5.2.4"
+        "webpack-dev-server": "5.2.5"
       }
     },
     "client/node_modules/@types/react": {
@@ -150,16 +150,6 @@
       "dependencies": {
         "tunnel": "^0.0.6",
         "undici": "^6.23.0"
-      }
-    },
-    "node_modules/@actions/http-client/node_modules/undici": {
-      "version": "6.25.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-6.25.0.tgz",
-      "integrity": "sha512-ZgpWDC5gmNiuY9CnLVXEH8rl50xhRCuLNA97fAUnKi8RRuV4E6KG31pDTsLVUKnohJE0I3XDrTeEydAXRw47xg==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">=18.17"
       }
     },
     "node_modules/@actions/io": {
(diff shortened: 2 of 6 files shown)


steilerDev:
**[product-architect]** Architecture & code-quality review

## Verdict: Approve

This is a clean, well-scoped security dependency update. The override strategy is the correct mechanism for this monorepo, all four targets resolve to single dev-only versions, and the lockfile is internally consistent.

---

### 1. Override strategy — appropriate ✅

Root `overrides` is the right tool here. In npm workspaces, transitive dev deps (`undici` via `@actions/http-client`, `uuid` via `sockjs`, `http-proxy-middleware` via `webpack-dev-server`) cannot be bumped by editing a leaf `package.json` — a root override is the only way to force a patched version across the hoisted tree. This matches the existing pattern already in `package.json` (`minimatch`, `picomatch`, `serialize-javascript`, `js-yaml`).

I verified the resulting lockfile collapses each target to exactly one resolved version, with no stale vulnerable major left behind:
- `undici` → **7.28.0** only (the previously-nested `@actions/http-client/node_modules/undici@6.25.0` is gone — correctly hoisted)
- `http-proxy-middleware` → **3.0.7** only
- `uuid` → **11.1.1** only
- `webpack-dev-server` → **5.2.5** only

### 2. Scope — dev-only, zero production risk ✅

Confirmed all four are `"dev": true` in the lockfile and reachable only from dev tooling (`webpack-dev-server`, `sockjs`, `@actions/http-client`). None appear in `server/package.json` production dependencies (`fastify`, `better-sqlite3`, `drizzle-orm`, `sharp`, `openid-client`, `tar`, etc.), so they are not present in the runtime Docker image. The production attack surface is unchanged.

### 3. Lockfile — reasonable ✅

The diff is consistent: leaf `client/package.json` bump (`5.2.4`→`5.2.5`) plus the four overrides, with the lockfile reflecting both. The removal of the duplicate nested `undici` entry and the `is-plain-obj@3` → `is-plain-object@5` swap (hpm v3's dependency change) are expected fall-out of the major bump. The three Prettier-reformatted server files are line-length-only with no logic change — no architectural concern.

### 4. Compatibility / forward-upgrade notes (non-blocking)

Two overrides force **major** version jumps past what the consumer declares — this is intentional with `overrides`, but worth recording:

- **`http-proxy-middleware` 2.0.9 → 3.0.7 (major):** `webpack-dev-server@5.2.5` still declares `"http-proxy-middleware": "^2.0.9"` in its manifest; the override forces v3 beyond that range. hpm v3's middleware API is compatible with wds's usage in practice, but this is the one item to keep an eye on if `npm run dev` proxying ever misbehaves.
- **`uuid` 8 → 11 (3 majors)** via `sockjs`, and **`undici` 6 → 7** via `@actions/http-client`: both are dev-tooling consumers; uuid's named-export change (v7+) is transparent to sockjs's usage.

**Forward path:** these overrides are a temporary forcing function. They should be removed once `webpack-dev-server` / `sockjs` / `@actions/http-client` ship releases that depend on the patched versions natively, to avoid silently pinning the whole tree. Recommend a tracking note so the overrides don't outlive their purpose.

### 5. CI note (informational, non-blocking for beta)

The required gate for `beta` — **Quality Gates — passes** (Docker build, smoke, unit, coverage all green). `E2E Shard 4/16` failed, but E2E runs against the **built Docker image** (Fastify-served static), not webpack-dev-server, so these dev-only overrides cannot affect E2E runtime. With the Docker build and the other 15 shards green, this reads as an unrelated flake rather than a regression from this change. Worth a re-run to confirm, but it is not a blocker for this PR's required gate.

---

No architecture, schema, or API-contract surfaces are touched — no wiki updates required. Approving from an architecture standpoint.

Issue #1786, opened 2026-06-25
Cost Breakdown: payment-status (Cost Basis) filter — Paid / Outstanding / All


steilerDev:
**As a** homeowner managing my building budget, **I want** a payment-status filter on the Cost Breakdown table **so that** I can quickly see how much money I have already spent versus how much is still left to be paid.

**Parent Epic**: _None — standalone story (budget/invoice backlog; related epics closed)_
**Priority**: Should Have

## Description

Add a **payment-status dropdown filter** ("Cost Basis") to the Cost Breakdown table on the Budget Overview page. It changes the cost basis used by every cost column and summary row, answering two common questions:

- "How much money have we already spent?" → **Paid**
- "How much is still left to be paid?" → **Outstanding**

### Dropdown options

- **All** (default): current behavior — projected costs for uninvoiced lines, actual invoiced amounts for invoiced lines.
- **Paid**: shows only paid+claimed portions of each budget line (deposit-aware, using `computeDepositAwareAggregates`). Uninvoiced lines contribute €0.
- **Outstanding**: shows pending invoice amounts for invoiced lines + projected planned amounts for uninvoiced lines (the full "still owed" picture, also deposit-aware for the invoiced portion).

### Technical context

- Invoice statuses: `pending | paid | claimed | quotation`. Deposit statuses: `pending | paid | claimed`.
- `computeDepositAwareAggregates` in `server/src/services/shared/depositAggregateUtils.ts` already computes `actualCostPaid` (paid+claimed, deposit-aware).
- The breakdown service at `server/src/services/budgetBreakdownService.ts` currently only exposes `actualCost` per budget line — it needs `actualCostPaid` added at the line, item, area, and totals level.
- The frontend `CostBreakdownTable` has an existing `PerspectiveToggle` (min/avg/max). The new dropdown is a **separate control** — it changes the cost basis, not the uncertainty range, and should render in the same area, labeled "Cost Basis" or similar.
- URL state should persist the selection (mirroring the existing `?deselectedSources=` param) via `?paymentStatus=paid|outstanding` (omit the param for "All").

## Acceptance Criteria

- [ ] AC1: A "Cost Basis" dropdown appears on the Cost Breakdown table with options All / Paid / Outstanding.
- [ ] AC2: Selecting "Paid" switches all cost columns to show only paid+claimed amounts (deposit-aware). Lines with no paid amount show €0.
[...]

PR #1788, opened 2026-06-26
feat(budget): add payment-status cost basis filter to cost breakdown (#1786)

steilerDev:
## Summary

- Adds "Cost Basis" dropdown (All / Paid / Outstanding) to the Cost Breakdown table, separate from the existing min/avg/max perspective toggle
- Backend computes deposit-aware `actualCostPaid` and `actualCostPending` on all breakdown entities using `computeDepositAwareAggregates`
- Frontend `resolveBudgetLineCost` / `resolveAggregateCost` helpers switch displayed cost based on selection; URL state persists via `?paymentStatus=paid|outstanding`

Fixes #1786

## Test plan

- [ ] Unit tests pass (95%+ coverage)
- [ ] Integration tests pass
- [ ] E2E cost-basis spec covers paid/outstanding filter flows across all viewports

Co-Authored-By: Claude dev-team-lead (claude-sonnet-4-6) <noreply@anthropic.com>


steilerDev:
**[security-engineer]**

## Security Review — PR #1788 (Issue #1786: Cost Basis payment-status filter)

**Verdict: APPROVED — no blocking findings.**

---

### Checklist

- [x] No SQL/command/XSS injection vectors in new code
- [x] Authentication/authorization enforced on all new endpoints (no new endpoints added)
- [x] No sensitive data exposed in logs, errors, or client responses
- [x] User input validated and sanitized at API boundaries
- [x] New dependencies: none added
- [x] Error responses do not leak internal details

---

### SQL Injection Analysis — `budgetBreakdownService.ts`

Both new queries (Query B for WI budget lines, Query D for HI budget lines) use Drizzle ORM `sql` tagged template literals and contain **zero user-controlled input**:

```sql
SELECT ibl.work_item_budget_id AS budgetLineId, ...
FROM invoice_budget_lines ibl
INNER JOIN invoices i ON i.id = ibl.invoice_id
LEFT JOIN invoice_deposits id ON id.invoice_id = i.id
WHERE ibl.work_item_budget_id IS NOT NULL
```

These are fully static queries — no runtime string interpolation, no parameterized values from user input. Row-level grouping and aggregation is performed in-process on the TypeScript side using `computeDepositAwareAggregates`. There is no injection surface.

### Input Validation — `paymentStatus` URL Parameter

The frontend applies a strict allowlist at the point of consumption in `BudgetOverviewPage.tsx`:

```typescript
const raw = searchParams.get('paymentStatus');
if (raw === 'paid' || raw === 'outstanding') return raw;
return 'all';  // default covers null, empty string, and any tampered value
```

Any value outside the three-item allowlist is silently clamped to `'all'`. The parameter never reaches the server — it is a pure client-side display switch that selects between the three amounts already returned by `GET /api/budget/breakdown`. This design is correct from a security perspective: keeping the filter client-side eliminates any server-side injection surface entirely.

### Data Exposure

The new `actualCostPaid` and `actualCostPending` fields are computed from invoice and deposit records already within the authenticated user's data scope. The existing RBAC session check on `GET /api/budget/breakdown` covers these fields without change. No new data categories or cross-tenant paths are introduced.

### `depositAggregateUtils.ts`

Pure arithmetic on in-memory data already retrieved from the DB. The `targetStatus: string` parameter in `computeStatusContribution` is called from service-layer code only, never from user input, and is used in JavaScript equality comparisons, not SQL string construction. No injection surface.

### `CostBasisSelect` Component

The `onChange` handler uses a TypeScript `as` assertion (`e.target.value as 'all' | 'paid' | 'outstanding'`) which does not provide runtime enforcement. The actual safety net is the `useMemo` allowlist in `BudgetOverviewPage.tsx` cited above. Even if the DOM were tampered to inject an arbitrary string, the value never reaches the server and is overwritten on the next render by the allowlist clamp. No exploit path.

---

No findings to document on the Security Audit wiki page. The implementation maintains the established security baseline for this codebase.

steilerDev:
**[product-architect]** Architecture / contract / quality review — **APPROVE** (medium & low findings noted for refinement, none blocking)

## Verified
- **Shared types** — `actualCostPaid` / `actualCostPending` added consistently to all six breakdown shapes (`BreakdownBudgetLine`, `BreakdownWorkItem`, `BreakdownArea`, `BreakdownHouseholdItem`, `BreakdownTotals`, `BudgetSourceSummaryBreakdown`) with JSDoc explaining semantics. snake_case→camelCase naming convention respected. No schema/migration changes (fields are computed, not persisted) — correct.
- **Deposit-aware aggregation** — Reuses the established `computeDepositAwareAggregates` helper (ADR-029 pattern, shared with `budgetSourceService`/`budgetOverviewService`). The switch from `GROUP BY … SUM()` to a `LEFT JOIN invoice_deposits` + in-memory aggregation is safe: the helper deduplicates by `ibl_id`, so the join fan-out (one row per deposit) does not double-count itemized amounts. `actualCostPending = actualCost - actualCostPaid` is computed identically at every rollup level (line → item → area → totals → per-source). Confirmed correct.
- **Per-source figures stay unfiltered** — `sourceProjectedMap` now carries `actualCost/Paid/Pending` computed from the full unfiltered line set, matching the existing `projectedMin/Max` convention and documented in the wiki. Scenario 10 test confirms a deselected source still reports full actuals.
- **API Contract wiki** — Updated with the new fields and explicit semantics (paid+claimed definition, deposit-proportional split note, "per-source always unfiltered" caveat). Contract and types match.
- **Client-side filtering decision** — Returning all three cost variants from the server and switching in the browser is the right call here: per-source actuals must remain unfiltered regardless of the active filter, the dataset is small (<5 users), and it avoids extra round-trips on every dropdown change. Approved.
- **URL state** — `?paymentStatus=paid|outstanding` (omitted for `all`) follows the same `useSearchParams` derive-from-URL pattern as `deselectedSources`. Consistent.
- **Test coverage** — 13 backend integration scenarios (non-invoiced, paid, claimed, pending, deposit splits incl. claimed deposits, item/area/global rollups, per-source, deselected-source, household items) + 8 client component tests (Scenarios 12–19) + E2E spec. Strong coverage of the deposit-aware paths.

## Findings (non-blocking — address in refinement)

**MEDIUM — Extract a named type for the filter union.** `'all' | 'paid' | 'outstanding'` is repeated 17× inline across `BudgetOverviewPage.tsx` and `CostBreakdownTable.tsx` (props, helpers, the `as` cast in the select handler). Introduce a single `type PaymentStatusFilter = 'all' | 'paid' | 'outstanding'` (shared, or a local export from the table module) so the option set has one source of truth and the `e.target.value as …` cast can't silently drift.

**MEDIUM — `isQuotation` semantics changed; add a mixed-invoice test.** The line-quotation rule moved from `MAX(i.status) = 'quotation'` to `rows.every(r => r.invoice_status === 'quotation')`. The new "all linked invoices are quotations" definition is clearly more correct than the old alphabetical-MAX heuristic, but it is a behavioral change for budget lines linked to multiple invoices, and I don't see a test covering a line with a mix of quotation + non-quotation invoices. Please confirm the change is intentional and add a scenario locking in the new behavior.

**LOW — Line-vs-aggregate drift for quotations in `outstanding` mode.** At line level, an `outstanding` quotation returns `quotedPerspectiveValue` (±5% of `actualCost`), while the aggregate path returns `actualCostPending` (= `actualCost` for a quotation, since paid=0). The sum of child line values can therefore differ slightly from the displayed aggregate. Likely acceptable, but worth a quick confirm with PO/UX that this is the intended display.

**LOW — Hardcoded `min-height: 44px` / `min-width: 120px` in the new CSS.** These are reasonable touch-target/sizing values and stylelint passed, but defer to the UX designer on whether a touch-target token should be used instead of the raw px.

No API-contract violations, no schema inconsistencies, and the deposit-aware pattern is applied correctly. Approving.

steilerDev:
**[ux-designer]** Design review for PR #1788 — Cost Basis (Payment Status) Filter

---

## Verdict: APPROVED

All new CSS classes and the `CostBasisSelect` component pass the full design system checklist.

---

### Token Adherence

All five new CSS classes use design tokens exclusively:

- `.controlBar` — `var(--spacing-6)` gap
- `.costBasisField` — `var(--spacing-2)` gap
- `.costBasisLabel` — `var(--font-size-sm)`, `var(--font-weight-medium)`, `var(--color-text-secondary)`
- `.costBasisSelect` — composes `select` from `shared.module.css` (correct reuse of the shared pattern), then overrides with `width: auto`, `min-height: 44px`, `min-width: 120px`
- `.costBasisSelectActive` — `border-color: var(--color-primary)`

No hardcoded hex values, magic spacing numbers (see Informational note below), or raw font sizes.

### Touch Target

`min-height: 44px` is explicitly set on `.costBasisSelect`. Meets the 44×44 px minimum.

### Dark Mode

Every color value flows through semantic CSS custom properties that are redefined in `[data-theme="dark"]`: `--color-text-secondary`, `--color-primary`, and all values inherited from the composed `.select` class (`--color-bg-primary`, `--color-text-primary`, `--color-border-strong`). No hardcoded light-only colors.

### Accessibility

- `<label htmlFor="cost-basis-select">` + `<select id="cost-basis-select">` — label is correctly associated via matching `htmlFor`/`id`.
- Native `<select>` element — keyboard-navigable without additional ARIA.
- `selectId` is a static constant (`'cost-basis-select'`). Verified there is only one `CostBreakdownTable` rendered per page in `BudgetOverviewPage.tsx` (line 354), so no duplicate-ID risk.

### Visual Consistency

- Composes from `shared.module.css` `.select` rather than re-implementing select styles — correct pattern.
- `.costBasisLabel` styling matches the established form label pattern used elsewhere.
- Placed in the existing `.controlBar` flex wrapper alongside `PerspectiveToggle` with a `var(--spacing-6)` gap.

### Active State

`border-color: var(--color-primary)` applied via `.costBasisSelectActive` when `value !== 'all'` gives a clear visual indicator that the filter is non-default. The border-color cascade is correct: `.costBasisSelectActive` appears after `.select` in the stylesheet so it wins the cascade without needing `!important`.

Focus state on top of active state: both `focused` and `active-not-focused` produce `--color-primary` border; `focused` additionally adds `--shadow-focus-subtle` (the glow ring from `.select:focus-visible`). The visual distinction is appropriate.

### Print

`.controlBar` is included in the existing `@media print` `display: none !important` rule, correctly hiding both the PerspectiveToggle and the Cost Basis select from print output.

### Responsive

Mobile media query (`max-width: 767px`) stacks `.costBasisField` vertically (`flex-direction: column; align-items: flex-start`) and sets `.costBasisSelect { width: 100% }`. Correct behavior.

### Informational (non-blocking)

`min-width: 120px` in `.costBasisSelect` has no direct spacing token equivalent (the scale tops out at `--spacing-16: 64px`). This is a semantic content-minimum-width for a form control rather than a spacing relationship, and is consistent with the pre-existing `min-width: 52px` on `.perspectiveSegment`. No action required.

steilerDev:
**[product-owner]** Requirements & Acceptance-Criteria review of #1786

**Verdict: APPROVED** (with one documented spec deviation + minor non-blocking notes)

I validated the PR line-by-line against the 13 acceptance criteria on Issue #1786.

### Acceptance Criteria

| AC | Status | Evidence |
| --- | --- | --- |
| AC1 — Cost Basis dropdown (All/Paid/Outstanding) | PASS | `CostBasisSelect` rendered in `.controlBar` (CostBreakdownTable.tsx:1038); E2E "Cost Basis control bar visible". |
| AC2 — Paid = paid+claimed (deposit-aware), non-invoiced €0 | PASS | `resolveBudgetLineCost` returns `actualCostPaid` for paid; aggregates return `actualCostPaid`. Backend `computeDepositAwareAggregates` feeds `actualCostPaid`. E2E covers invoiced €5,000 and non-invoiced €0. |
| AC3 — Outstanding = pending invoiced + projected uninvoiced (deposit-aware) | PASS | `resolveBudgetLineCost` / `resolveAggregateCost` (pending + projected portion). E2E covers €2,000 pending and €3,000 projected. |
| AC4 — All restores default | PASS | Default branch preserves prior behavior; E2E "switch back to All". |
| AC5 — URL `?paymentStatus=paid|outstanding` (absent = All) | PASS | `handlePaymentStatusChange` adds/removes param; invalid value falls back to All. E2E for add/remove/restore/invalid. |
| AC6 — Change triggers refetch (debounce + AbortController) | PASS (redundant) | `paymentStatus` is in the refetch `useEffect` deps (BudgetOverviewPage.tsx:163). See note 2 — the refetch is functionally redundant under the chosen client-side design. |
| AC7 — Perspective stays active; controls projected uncertainty in Outstanding | PASS | `resolveAggregateCost` applies perspective to the projected portion; E2E verifies Min vs Avg changes the Outstanding projected line. |
| AC8 — Available Funds / Remaining Budget reflect filtered totals | PASS | `totalRawProjected` is filter-aware; Remaining Budget = `filteredAvailableFunds − totalRawProjected (+ payback)`. Available Funds itself is correctly left unfiltered (funds don't depend on payment state). |
| AC9 — Per-source rows filter-aware | PASS | Per-source `allocatedCost` uses `resolveAggregateCost(..., paymentStatus)`. (Subsidy payback stays projection-only — reasonable, see note 3.) |
| AC10 — EN keys added, DE provided | PASS | `overview.costBreakdown.costBasis.{label,all,paid,outstanding}` present in both `en` and `de`. |
| AC11 — Backend unit coverage ≥95% | PASS | New `budgetBreakdownService.paid.test.ts` covers paid/pending aggregation. |
| AC12 — Integration tests on `?paymentStatus=` query param on `GET /api/budget/breakdown` | **DEVIATION** | No server-side `?paymentStatus=` param exists — filtering is done client-side from additive backend fields (`actualCostPaid` / `actualCostPending`). See deviation below. |
| AC13 — E2E for dropdown + URL persistence | PASS | `budget-cost-basis.spec.ts` — comprehensive across desktop/tablet/mobile. |

### Spec deviation (accepted)

AC6/AC12 were written assuming **server-side filtering** (a `?paymentStatus=` query param the breakdown endpoint would honor, integration-tested). The implementation instead makes the backend change **purely additive** — it returns `actualCost`, `actualCostPaid`, and `actualCostPending` on every line/item/area/totals/source — and the frontend switches which value it displays. This is a sound, arguably superior choice: all data is already present, so switching is instant with no round-trip, and existing `actualCost` consumers are untouched (exactly as the story Notes intended). The behavioral intent of AC12 is still covered — by `budgetBreakdownService.paid.test.ts` (unit) plus the full E2E suite. I'm accepting this deviation; no functional AC is unmet. Recording it here so the divergence from the written ACs is on the record.

### Non-blocking notes (cleanup, not merge blockers)

1. **Dead variable** — `_sum` (CostBreakdownTable.tsx:914) duplicates the inline Remaining-Budget computation at ~1386–1392 and is unused. Remove it.
2. **Redundant refetch** — because filtering is client-side and `paymentStatus` is never sent to the server, the refetch fired on every dropdown change is a wasteful no-op network call. Consider dropping `paymentStatus` from the refetch `useEffect` deps (the data needed for all three modes is already in the loaded breakdown). This satisfies AC6's letter but not its purpose.
3. **Per-source subsidy payback** stays perspective-only (not payment-filtered) while allocated cost is filter-aware — acceptable, since subsidy payback is inherently a projection rather than a paid/pending invoice value.

### Process note

Only a `security-engineer` comment is present on this PR; I don't see a recorded architecture or QA review. Given the client-side-vs-server-side architectural deviation above, please confirm `product-architect` sign-off before promotion.

Functionally complete and well-tested. Approving for merge to `beta`.

